PR #4, opened 2021-05-18
Attempt to fix flaky tests

meziantou:


NGitLab.Tests/MergeRequest/MergeRequestClientTests.cs
```diff
@@ -25,17 +26,20 @@ public async Task Test_merge_request_api()
             await GitLabTestContext.RetryUntilAsync(
                     () => mergeRequestClient[mergeRequest.Iid],
                     mr => string.Equals(mr.MergeStatus, "can_be_merged", StringComparison.Ordinal),
-                    TimeSpan.FromSeconds(120))
+                    TimeSpan.FromSeconds(240))
```
louis-z: Retry for up to 4 minutes, really? I thought 2 minutes was already a bit long... unless perhaps we're waiting for the GitLab server to be up and running?

NGitLab.Tests/MergeRequest/MergeRequestClientTests.cs
```diff
@@ -25,17 +26,20 @@ public async Task Test_merge_request_api()
...
                 .ConfigureAwait(false);
 
+            TestContext.Out.WriteLine("MR is ready to be merged");
             AcceptMergeRequest(mergeRequestClient, mergeRequest);
-
-            await GitLabTestContext.RetryUntilAsync(
-                    () => context.Client.GetRepository(project.Id).Branches.All.FirstOrDefault(x => string.Equals(x.Name, mergeRequest.SourceBranch, StringComparison.Ordinal)),
-                    branch => branch == null,
```
louis-z, line 35: Last week I investigated a bit more this flaky test. It seems that sometimes GitLab does not apply all the following in their calling order:

https://github.com/ubisoft/NGitLab/blob/f6b8b1363c3af86c1841f564859dc78d38463846/NGitLab.Tests/Docker/GitLabTestContext.cs#L185-L187

In such cases, "branch" would end up being a protected branch and so the automatic deletion would fail. We should look further into that.

NGitLab.Tests/MergeRequest/MergeRequestClientTests.cs
meziantou: I think I can revert this change as it doesn't help

NGitLab.Tests/MergeRequest/MergeRequestClientTests.cs
meziantou: GitLab seems to add more an more asynchronous operations. Plus their GitLab container being very flaky, so it's not easy to understand the actual failure reason... Adding a delay between Accept and GetBranches seems to help, but I don't know how many seconds is safe. Looking at the website, it seems the branch is deleted almost immediately, but the API still returns the deleted branch... 

Also, I cannot reproduce the issue on my machine when running just this test. I can sometimes reproduce it when running all tests from this class.

At the moment, I think we should just remove this part of the test as it is too flaky.

meziantou: Indeed, GitLab seems to change the default branch and so, protect the wrong branch...

image

jontdelorme (approved):

PR #6, opened 2021-05-20
Fix mock - main as initial branch

Toa741:
Following the change @ https://github.com/ubisoft/NGitLab/commit/b2566a1ec8cd0dec90f6147665a291688800998d, mocked repository was still initialized on "master" branch.

As libgit2sharp doesn't have the option to provide an initial branch, replaced the Init action with a direct git call

### DIFF
--- a/NGitLab.Mock/Project.cs
+++ b/NGitLab.Mock/Project.cs
@@ -38,7 +38,7 @@ namespace NGitLab.Mock
 
         public string Description { get; set; }
 
-        public string DefaultBranch { get; set; } = GitLabServer.DefaultBranch;
+        public string DefaultBranch { get; set; } = GitLabServer.DefaultBranchName;
 
         public string WebUrl => Server.MakeUrl(PathWithNamespace);
 
--- a/NGitLab.Mock/Project.cs
+++ b/NGitLab.Mock/Project.cs
@@ -38,7 +38,7 @@ namespace NGitLab.Mock
 
         public string Description { get; set; }
 
-        public string DefaultBranch { get; set; } = "main";
+        public string DefaultBranch { get; set; } = Repository.DefaultBranch
 
         public string WebUrl => Server.MakeUrl(PathWithNamespace);
 
--- a/NGitLab.Mock/Project.cs
+++ b/NGitLab.Mock/Project.cs
@@ -38,7 +38,7 @@ namespace NGitLab.Mock
 
         public string Description { get; set; }
 
-        public string DefaultBranch { get; set; } = Repository.DefaultBranch
+        public string DefaultBranch { get; set; } = GitLabServer.DefaultBranch;
 
         public string WebUrl => Server.MakeUrl(PathWithNamespace);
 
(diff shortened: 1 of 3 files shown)


NGitLab.Mock/Repository.cs
```diff
@@ -50,7 +51,11 @@ private LibGit2Sharp.Repository GetGitRepository()
                         var directory = TemporaryDirectory.Create();
                         if (Project.ForkedFrom == null)
                         {
-                            LibGit2Sharp.Repository.Init(directory.FullPath);
+                            // libgit2sharp cannot init with a branch other than master
+                            using var process = Process.Start("git", $"init --initial-branch {DefaultBranch} \"{directory.FullPath}\"");
+                            process.WaitForExit();
+                            if (process.ExitCode != 0)
+                                throw new GitLabException("Cannot init repository");
```
louis-z: Please add `directory.FullPath` to the exception message.

louis-z (changes requested):

Toa741:
We use a CommandLineHelpers to escape arguments in another project. Should I add it in the Mock project for having a better way of escaping?

louis-z (approved):

louis-z:
> We use a CommandLineHelpers to escape arguments in another project. Should I add it in the Mock project for having a better way of escaping?

NGitLab.Mock/Repository.cs
```diff
@@ -12,6 +12,7 @@ namespace NGitLab.Mock
 {
     public sealed class Repository : GitLabObject, IDisposable
     {
+        public const string DefaultBranch = "main";
```
louis-z: We should define the fallback default branch only once in the codebase. "main" is also specified here:

https://github.com/ubisoft/NGitLab/blob/cf0a15949dfc582607c0d58038f89bb69eb3c4a9/NGitLab.Mock/Project.cs#L41

Maybe replace that by
```cs
public sealed class Project : GitLabObject
{
    public string DefaultBranch { get; set; } = Repository.DefaultBranch;
}
```
?

meziantou (changes requested):
I think you should add a `GitLabServer.DefaultBranchName {get;set;} = "main"` and use this value everywhere.

NGitLab.Mock/Repository.cs
```diff
@@ -52,7 +51,7 @@ private LibGit2Sharp.Repository GetGitRepository()
                         if (Project.ForkedFrom == null)
                         {
                             // libgit2sharp cannot init with a branch other than master
-                            using var process = Process.Start("git", $"init --initial-branch {DefaultBranch} \"{directory.FullPath}\"");
+                            using var process = Process.Start("git", $"init --initial-branch {GitLabServer.DefaultBranch} \"{directory.FullPath}\"");
```
meziantou: You should use `Project.DefaultBranch`

NGitLab.Mock/Project.cs
```diff
@@ -38,7 +38,7 @@ public Project(string name)
 
         public string Description { get; set; }
 
-        public string DefaultBranch { get; set; } = Repository.DefaultBranch
+        public string DefaultBranch { get; set; } = GitLabServer.DefaultBranch;
```
meziantou: The logic should be more complex. When the server (Parent) is set for this instance, it should set the default branch value unless there is an explicit value.

meziantou (approved):

PR #7, opened 2021-05-20
Improve internal retry scheme

louis-z:
- Disable internal retry scheme for non-idempotent HTTP requests
- Convert `FunctionRetryExtensionsTests.Retry` from recursive to iterative
- Log debug info about internal retries during tests

### DIFF
--- a/NGitLab/Extensions/FunctionRetryExtensions.cs
+++ b/NGitLab/Extensions/FunctionRetryExtensions.cs
@@ -22,7 +22,7 @@ namespace NGitLab.Extensions
                 {
                     return action();
                 }
-                catch (Exception ex) when (predicate(ex, retriesLeft))
+                catch (Exception ex) when (retriesLeft > 0 && predicate(ex, retriesLeft))
                 {
                     Logger?.Invoke($"{ex.Message} -> Internal Retry in {waitTime.TotalMilliseconds} ms ({maxRetryCount - retriesLeft + 1} of {maxRetryCount})...");
 
--- a/NGitLab/Extensions/FunctionRetryExtensions.cs
+++ b/NGitLab/Extensions/FunctionRetryExtensions.cs
@@ -8,6 +8,8 @@ namespace NGitLab.Extensions
     /// </summary>
     internal static class FunctionRetryExtensions
     {
+        public static Action<string> Logger { get; set; }
+
         /// <summary>
         /// Do a retry a number of time on the received action if it fails
         /// </summary>
@@ -19,6 +21,8 @@ namespace NGitLab.Extensions
             }
             catch (Exception ex) when (retryWhen(ex, retryNumber))
             {
+                Logger?.Invoke($"{ex.Message} -> Internal Retry ({retryNumber - 1} attempts left)...");
+
                 Thread.Sleep(interval);
 
                 var nextInterval = isIncremental ? interval.Add(interval) : interval;
--- a/NGitLab/Extensions/FunctionRetryExtensions.cs
+++ b/NGitLab/Extensions/FunctionRetryExtensions.cs
@@ -13,20 +13,26 @@ namespace NGitLab.Extensions
         /// <summary>
         /// Do a retry a number of time on the received action if it fails
         /// </summary>
-        public static T Retry<T>(this Func<T> action, Func<Exception, int, bool> retryWhen, TimeSpan interval, int retryNumber, bool isIncremental)
+        public static T Retry<T>(this Func<T> action, Func<Exception, int, bool> predicate, TimeSpan waitTime, int maxRetryCount, bool useExponentialBackoff)
         {
-            try
+            var retriesLeft = maxRetryCount;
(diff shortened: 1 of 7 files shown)


louis-z:
Some tests are still randomly failing... Let's see where this PR takes us.

NGitLab/Impl/HttpRequestor.GitLabRequest.cs
```diff
@@ -58,6 +58,11 @@ public GitLabRequest(Uri url, MethodType method, object data, string apiToken, s
 
             public WebResponse GetResponse(RequestOptions options)
             {
+                // For any requests that are POTENTIALLY NOT idempotent, send only once
```
meziantou: I think we should update the RequestOptions with a new method 

````
public virtual bool ShouldRetry(Exception ex, WebResponse response, int retryNumber)
````

We can change the default ShouldRetry implementation to skip non-GET methods. But we still allow people to retry POST method if they want to.

Comment (louis-z) [reply NGitLab/Impl/HttpRequestor.GitLabRequest.cs]: I like the idea of giving the user the possibility to choose the old behavior. However, I'm not sure that adding a new ShouldRetry (taking in an additional WebResponse) is the best approach, for at least a few reasons:
- The Retry\<T> method is generic and therefore does not know about WebResponse. We could probably try to cast T to WebResponse and work with that, if it weren't for the other reasons.
- In the said method, the WebResponse is actually returned by calling action(), which actually triggers the retry when it throws. At that point there is no returned WebResponse to pass in to ShouldRetry.
- I think that a simpler approach is just to offer the user a bool to indicate they want the old behavior.

Comment (louis-z) [reply NGitLab/Impl/HttpRequestor.GitLabRequest.cs]: Ok, so I revisited the code as follows:

 - GitLabException now stores the HTTP request method, if any, that triggered the exception
- RequestOptions.ShouldRetry was adapted so that
  - HTTP request methods that are NOT safe will not be retried
  - We use the same Transient HTTP StatusCodes as Polly, i.e. >= 500 or == 408

File: NGitLab/RequestOptions.cs [hunk 2]
```diff
@@ -39,7 +39,24 @@ public RequestOptions(int retryCount, TimeSpan retryInterval, bool isIncremental
...
+
+            if (!(ex is GitLabException gitLabException))
+                return false;
+
+            // For requests that are potentially NOT Safe/Idempotent, do not retry
+            // See https://developer.mozilla.org/en-US/docs/Glossary/Safe/HTTP
+            // If there is no HTTP request method specified, carry on to the next condition below.
+            if (gitLabException.MethodType.HasValue &&
+                gitLabException.MethodType != Impl.MethodType.Get &&
+                gitLabException.MethodType != Impl.MethodType.Head)
```
meziantou:
```suggestion
                gitLabException.MethodType != Impl.MethodType.Head &&
                gitLabException.MethodType != Impl.MethodType.Options)
```

NGitLab/Extensions/FunctionRetryExtensions.cs
```diff
@@ -8,22 +8,32 @@ namespace NGitLab.Extensions
...
-            catch (Exception ex) when (retryWhen(ex, retryNumber))
+            var retriesLeft = maxRetryCount;
+            while (true)
             {
-                Thread.Sleep(interval);
+                try
+                {
+                    return action();
+                }
+                catch (Exception ex) when (predicate(ex, retriesLeft))
```
meziantou: We should force the handling of `retriesLeft`

```suggestion
                catch (Exception ex) when (retriesLeft > 0 && predicate(ex, retriesLeft))
```

NGitLab/Extensions/FunctionRetryExtensions.cs
```diff
@@ -8,22 +8,32 @@ namespace NGitLab.Extensions
...
+                {
+                    Logger?.Invoke($"{ex.Message} -> Internal Retry in {waitTime.TotalMilliseconds} ms ({maxRetryCount - retriesLeft + 1} of {maxRetryCount})...");
+
+                    Thread.Sleep(waitTime);
+
+                    if (useExponentialBackoff)
+                        waitTime = waitTime.Add(waitTime);
```
meziantou, line 33:
```suggestion
                    {
                        waitTime = waitTime.Add(waitTime);
                    }                        
```

NGitLab/Extensions/FunctionRetryExtensions.cs
louis-z: The `predicate` should handle it, as before. Otherwise what's the point in passing in `retriesLeft` to the predicate?

If a user overrides the default predicate and forgets to take `retriesLeft` into account, well I say tough luck. 😁

meziantou: > Otherwise what's the point in passing in retriesLeft to the predicate?

logging/informative, but I think we should enforce it to avoid bad implementations

PR #9, opened 2021-05-25
Use symbolic-ref instead on initial-branch for Git repo init in Mock

Toa741:
- `--initial-branch` is only available on recent version of Git
- stable version in debian 10 is older and doesn't have the option
- Add errors details for easier debug

### DIFF
--- a/NGitLab.Mock/Repository.cs
+++ b/NGitLab.Mock/Repository.cs
@@ -50,11 +50,27 @@ namespace NGitLab.Mock
                         var directory = TemporaryDirectory.Create();
                         if (Project.ForkedFrom == null)
                         {
+                            LibGit2Sharp.Repository.Init(directory.FullPath);
+
                             // libgit2sharp cannot init with a branch other than master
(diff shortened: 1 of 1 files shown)


NGitLab.Mock/Repository.cs
```diff
@@ -50,11 +50,27 @@ private LibGit2Sharp.Repository GetGitRepository()
...
+
                             // libgit2sharp cannot init with a branch other than master
-                            using var process = Process.Start("git", $"init --initial-branch \"{Project.DefaultBranch}\" \"{directory.FullPath}\"");
+                            // Use symbolic-ref to keep the code compatible with older version of git
+                            using var process = new Process
+                            {
+                                StartInfo = new ProcessStartInfo
+                                {
+                                    FileName = "git",
+                                    Arguments = $"-C \"{directory.FullPath}\" symbolic-ref HEAD refs/heads/{Project.DefaultBranch}",
```
meziantou: You can use `WorkingDirectory` instead of `-C` to make it easier to understand.

Also you should quote the last parameter

Toa741: Indeed :).
Added the working dir & last arg quoted

PR #13, opened 2021-05-31
Add functionnality to retrieve test reports from a pipeline

noctur4:
Add functionnality to retrieve test reports from a pipeline as the test reports is accessible via the information of a pipeline when the tests results are in Junit.

It also means that with that functionnality we can retrieve the stats all the type of tests that use a junit report format and pull metrics.

It's usefull for anybody using GitLab and for the TG

doc:
https://docs.gitlab.com/ee/api/pipelines.html#get-a-pipelines-test-report

examples of a test report call on the GitLab API : 
https://gitlab-ncsa.ubisoft.org/api/v4/projects/51299/pipelines/4080199/test_report
https://gitlab-ncsa.ubisoft.org/api/v4/projects/10207/pipelines/4079763/test_report

### DIFF
--- a/NGitLab/Models/TestReport.cs
+++ b/NGitLab/Models/TestReport.cs
@@ -4,7 +4,7 @@ using System.Runtime.Serialization;
 namespace NGitLab.Models
 {
     [DataContract]
-    public class TestReports
+    public class TestReport
     {
         [DataMember(Name = "total_time")]
         public int TotalTime { get; set; }
--- a/NGitLab/Models/TestReport.cs
+++ b/NGitLab/Models/TestReport.cs
@@ -25,6 +25,6 @@ namespace NGitLab.Models
         public int ErrorCount { get; set; }
 
         [DataMember(Name = "test_suites")]
-        public TestSuites[] TestSuites { get; set; }
+        public IReadOnlyCollection<TestSuites> TestSuites { get; set; }
     }
 }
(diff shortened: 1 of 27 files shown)


noctur4:
@Toa741 I'm ready to be reviewed.

NGitLab/Models/TestReport.cs
```diff
@@ -0,0 +1,30 @@
...
+        public int FailedCount { get; set; }
+
+        [DataMember(Name = "skipped_count")]
+        public int SkippedCount { get; set; }
+
+        [DataMember(Name = "error_count")]
+        public int ErrorCount { get; set; }
+
+        [DataMember(Name = "test_suites")]
+        public TestSuites[] TestSuites { get; set; }
```
meziantou:
```suggestion
        public IReadOnlyCollection<TestSuites> TestSuites { get; set; }
```

NGitLab/Models/TestSuites.cs
```diff
@@ -0,0 +1,33 @@
...
+        public int FailedCount { get; set; }
+
+        [DataMember(Name = "skipped_count")]
+        public int SkippedCount { get; set; }
+
+        [DataMember(Name = "error_count")]
+        public int ErrorCount { get; set; }
+
+        [DataMember(Name = "test_cases")]
+        public TestCases[] TestCases { get; set; }
```
meziantou:
```suggestion
        public IReadOnlyCollection<TestCases> TestCases { get; set; }
```

meziantou (changes requested):

meziantou (approved):

Toa741 (approved):

PR #19, opened 2021-06-09
Add additional info in Mock's Change

Toa741:


NGitLab.Mock/MergeRequest.cs
```diff
@@ -80,13 +80,12 @@ public MergeRequestChangeCollection Changes
         {
             get
             {
-                var tipCommit = SourceProject.Repository.GetBranchTipCommit(SourceBranch);
                 var stats = Project.Repository.GetBranchFullPatch(SourceBranch);
                 var changes = new MergeRequestChangeCollection(this);
                 foreach (var stat in stats)
                 {
                     var diff = stat.Patch.Substring(stat.Patch.IndexOf("@@", StringComparison.InvariantCulture));
```
meziantou:
```suggestion
                    var diff = stat.Patch.Substring(stat.Patch.IndexOf("@@", StringComparison.Ordinal));
```

NGitLab.Mock/MergeRequestChangeCollection.cs
```diff
@@ -10,11 +11,25 @@ public MergeRequestChangeCollection(GitLabObject parent)
         }
 
         public Change Add(string diff)
```
meziantou, line 14: Should we obsolete this method?

NGitLab.Mock/MergeRequestChangeCollection.cs
```diff
@@ -10,11 +11,25 @@ public MergeRequestChangeCollection(GitLabObject parent)
...
             if (diff is null)
                 throw new ArgumentNullException(nameof(diff));
 
-            var change = new Change() { Diff = diff };
+            var change = new Change
+            {
+                Diff = diff,
+                NewPath = newPath,
+                OldPath = oldPath,
+                DeletedFile = changeKind.HasValue && changeKind.Value == ChangeKind.Deleted,
```
louis-z: You can simplify the condition as follows (same for other 2)
```suggestion
                DeletedFile = changeKind.Value == ChangeKind.Deleted,
```

Toa741: changeKind is nullable, wouldn't it trigger an exception?

NGitLab.Mock/MergeRequestChangeCollection.cs
Toa741: Yes, would make things better in the end

NGitLab.Mock/MergeRequestChangeCollection.cs
meziantou: Nope, `T? item1 == T item2` is equivalent to `item1.HasValue && item1.Value == item2`

Toa741: Oh, I think it's just the suggestion given by louis:
`changeKind.Value == ChangeKind.Deleted` => `changeKind== ChangeKind.Deleted`

Will update that.

NGitLab.Mock/MergeRequest.cs
Toa741: Updated.

PR #20, opened 2021-06-11
Expose missing properties in Project mock

mogawinch:
- HttpUrl & SshUrl are now expose to use them in some tests

### DIFF
--- a/NGitLab.Mock/Project.cs
+++ b/NGitLab.Mock/Project.cs
@@ -57,6 +57,10 @@ namespace NGitLab.Mock
 
         public string WebUrl => Server.MakeUrl(PathWithNamespace);
 
+        public string SshUrl => Repository.FullPath;
+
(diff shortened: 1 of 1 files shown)


meziantou (commented):
Using a path instead of a url is misleading. What is your use case?

mogawinch:
@meziantou 
I'm creating some EndToEnds test in the worker with the mock of a GitLab server, so we need the SShUrl and HttpUrl to be compatible with a `git clone` command to run the ScanDependencies. That's why we use  the `Repository.FullPath`. 

It was actually already set to `Repository.FullPath` because we already have some tests that need to execute ScanDependencies in our tests (which run a `git clone`). On my side, I exposed it to do some setup with the generated value.

We could transform the Repository.FullPath or add comments to make it more obvious.

mogawinch:
> Using a path instead of a url is misleading. What is your use case?

PR #23, opened 2021-06-16
Added  pipeline getJobs with options for query

phdesroUbi:


NGitLab/IPipelineClient.cs
```diff
@@ -34,6 +34,11 @@ public interface IPipelineClient
         /// </summary>
         Job[] GetJobs(int pipelineId);
 
+        /// <summary>
+        /// Returns the jobs of a pipeline.
+        /// </summary>
+        Job[] GetJobs(PipelineGetJobsQuery query);
```
meziantou:
```suggestion
        IEnumerable<Job> GetJobs(PipelineGetJobsQuery query);
```

NGitLab/Models/PipelineGetJobsQuery.cs
```diff
@@ -0,0 +1,19 @@
...
+    [DataContract]
+    public class PipelineGetJobsQuery
+    {
+        [Required]
+        [DataMember(Name = "id")]
+        public int PipelineId { get; set; }
+
+        public string[] Scope { get; set; }
+
+        public bool IncludeRetried { get; set; }
```
meziantou:
```suggestion
        public bool? IncludeRetried { get; set; }
```

meziantou (changes requested):

NGitLab/Models/PipelineGetJobsQuery.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Runtime.Serialization;
+
+namespace NGitLab.Models
+{
+    [DataContract]
+    public class PipelineGetJobsQuery
```
louis-z: `PipelineJobQuery` would be more in line with the way we name query types

louis-z (changes requested):

meziantou (approved):

NGitLab/Models/PipelineGetJobsQuery.cs
louis-z: I suggested `PipelineJobQuery` (`Job`, singular), not `PipelineJobsQuery`. See other types:
- `EpicQuery`
- `IssueQuery`
- `JobQuery`
- `MergeRequestQuery`

The noun preceding Query is always singular.

phdesroUbi: Sorry, I thought that was because all of those things were only used in their respective client. Pipeline is singular because it's linked to the PipelineClient and Jobs plural because it's the jobs related to that pipeline.

I can make a quick pull-request if you want and make it singular.

louis-z: Please do

PR #30, opened 2021-06-23
Add Get user by query

Toa741:


NGitLab/Models/UserQuery.cs
```diff
@@ -0,0 +1,50 @@
+﻿namespace NGitLab
+{
+    public class UserQuery
+    {
+        /// <summary>
+        /// Specifies how many record per paging (Gitlab supports a maximum of 100 projects and defaults to 20).
```
louis-z:
```suggestion
        /// Specifies how many records per page (GitLab supports a maximum of 100 items per page and defaults to 20).
```

NGitLab/Models/UserQuery.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// </summary>
+        public bool? IsBlocked;
+
+        /// <summary>
+        /// If true, get only users that are external
+        /// </summary>
+        public bool? IsExternal;
+
+        /// <summary>
+        /// Exclude the extenral users
```
louis-z:
```suggestion
        /// Exclude external users
```

NGitLab/Models/UserQuery.cs
```diff
@@ -0,0 +1,50 @@
+﻿namespace NGitLab
+{
+    public class UserQuery
+    {
+        /// <summary>
+        /// Specifies how many record per paging (Gitlab supports a maximum of 100 projects and defaults to 20).
+        /// </summary>
+        public int? PerPage;
```
meziantou: Idem for other properties

```suggestion
        public int? PerPage { get; set;} 
```

NGitLab.Mock/UserCollection.cs
```diff
@@ -63,5 +63,58 @@ internal IEnumerable<User> SearchByUsername(string username)
         {
             return this.Where(user => string.Equals(user.UserName, username, StringComparison.OrdinalIgnoreCase));
         }
+
+        internal IEnumerable<User> Get(UserQuery query)
+        {
+            var users = this.AsQueryable();
+
+            if (query.IsActive.HasValue && query.IsActive.Value)
```
louis-z: You can make the condition simpler and clearer as follows:
```suggestion
            if (query.IsActive == true)
```

Same thing for other following conditions...

louis-z (changes requested):

NGitLab/Models/UserQuery.cs
Toa741: Updated

NGitLab/Models/UserQuery.cs
Toa741: Updated

NGitLab.Mock/UserCollection.cs
Toa741: Updated

NGitLab/Models/UserQuery.cs
Toa741: Updated. I also changed the other queries that had the same summary

louis-z (approved):

PR #31, opened 2021-06-23
Fix url generation when "search" is null

meziantou:


jairbubbles:
This one is begging for a unit test! 😊

PR #32, opened 2021-06-25
Move releases from TagClient to its own class

Bromvlieg:
With Gitlab 14 you cannot use tags api to create releases anymore.
This PR moves the releases from the TagClient to a new class ReleaseClient, implementing the same behavior as the other Clients.


jairbubbles:
Thanks for the contribution @Bromvlieg, this is very valuable as this is definitely something we identified on our side to allow migrating to GitLab 14.0.

We'll look more deeply at your changes but I feel like we should keep the old routes (maybe as obsolete) for now while users migrate to the new GitLab version.

Bromvlieg:
I'm not too familiar with NuGet, but perhaps there's a way to tag a mayor version for GitLab?
I noticed the docker image is set to a 13.x version rather than latest, so the tests are failing on that part.

Problem lies in Gitlab 14 throws a 500 internal server error when the `releaseDescription` field is used, I'm not sure what the best migration path would be for everyone. On our end it was a small change to use `client.Releases` rather than `client.Tags` so I don't think the impact will be problematic.

jairbubbles:
We could maintain two packages but it would be too much work and I don't think it's worth it.

We expect people to have their on-premises GitLab servers up to date. But in our case, it will take weeks before it gets updated. So we need to find a way to make both usages cohabit. 

I feel like the simpler for now would be only to add the new `ReleasesClient` and don't touch the old api / tests.

What do you think?

Bromvlieg:
That sounds like the easiest solution for now yeah.
I'll undo the tests and removals I did and leave the new class in.

We can add back the tests in a few months when most gitlabs are updated.
Would adding the `[depricated]` tags to the `releaseDescription` field work? then users also have a grace period on the library side at the same time.

jairbubbles:
Awesome!

> Would adding the [depricated] tags to the releaseDescription field work? then users also have a grace period on the library side at the same time.

Is the releases API properly supported in 13.x versions ? If so, yes it's a good idea to deprecate the old API with a clear message inviting user what to use instead. (Note: the attribute is called Obsolete: https://docs.microsoft.com/en-us/dotnet/api/system.obsoleteattribute?view=net-5.0)

NGitLab/Models/ReleaseInfo.cs
```diff
@@ -3,7 +3,7 @@
 namespace NGitLab.Models
 {
     [DataContract]
-    public class RealeaseInfo
+    public class ReleaseInfo
```
jairbubbles, line 6: Snake eyes!

NGitLab/Models/ReleaseCreate.cs
```diff
@@ -1,10 +1,28 @@
-﻿using System.Runtime.Serialization;
+﻿using System.ComponentModel.DataAnnotations;
+using System.Runtime.Serialization;
 
 namespace NGitLab.Models
 {
     [DataContract]
     public class ReleaseCreate
```
jairbubbles, line 7: As this class is shared with the old route, won't it break it? I feel like it would be safer to create new ones for the new routes.

NGitLab/ITagClient.cs
```diff
@@ -9,10 +9,10 @@ public interface ITagClient
...
         void Delete(string name);
 
-        IEnumerable<Tag> All { get; }
+        ReleaseInfo CreateRelease(string name, ReleaseCreate data);
 
-        RealeaseInfo CreateRelease(string name, ReleaseCreate data);
+        ReleaseInfo UpdateRelease(string name, ReleaseUpdate data);
 
-        RealeaseInfo UpdateRelease(string name, ReleaseUpdate data);
+        IEnumerable<Tag> All { get; }
```
jairbubbles, line 12: The method order was messed up. Any reason?

NGitLab/Impl/ReleaseClient.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System.Collections.Generic;
+using System.Net;
+using NGitLab.Models;
+
+namespace NGitLab.Impl
+{
+    public class ReleaseClient : IReleaseClient
```
jairbubbles, line 9: Maybe add a small comment and a link to the documentation?

https://docs.gitlab.com/ee/api/releases/

jairbubbles (commented):
I made small comments but overall the changes look good, thank you for your time!

Also it seems you have some analyzer issues (missing lines).

NGitLab/Impl/TagClient.cs
```diff
@@ -25,16 +25,17 @@ public void Delete(string name)
             _api.Delete().Stream($"{_tagsPath}/{WebUtility.UrlEncode(name)}", _ => { });
         }
 
-        public RealeaseInfo CreateRelease(string name, ReleaseCreate data)
+        public ReleaseInfo CreateRelease(string name, ReleaseCreate data)
```
jairbubbles: Theses methods should be deprecated too.

jairbubbles: After checking our code, we don't use those routes so I'm fine if we just remove them. That way we don't need to duplicate `ReleaseCreate` / `ReleaseUpdate`.

NGitLab/Models/ReleaseCreate.cs
jairbubbles: Not relevant anymore if we remove the old create / update methods.

NGitLab/Impl/TagClient.cs
jairbubbles: Like you did at the beginning... 😅 (but keep `ReleaseDescription` in `TagCreate`)

NGitLab/Models/ReleaseInfo.cs
Bromvlieg: Can't resist :) sooner or later it's going to be a breaking change.

NGitLab/Models/ReleaseCreate.cs
Bromvlieg: It can share it as it doesn't modify fields, if we add new ones it'll add extra migration later on

NGitLab/ITagClient.cs
Bromvlieg: No reason, will correct it in a commit

NGitLab/Impl/ReleaseClient.cs
Bromvlieg: the others don't have it, feels odd to add it for this one.

jairbubbles:
Hi @Bromvlieg, as we haven't heard from you for quite some days I was just checking to make sure this PR is still in your radar.

Bromvlieg:
Hey Jairbubbles. Yes! I had vacation last week :) Tomorrow I'll be spending a few hours on tooling where I will resume the PR.

NGitLab/ITagClient.cs
Bromvlieg: removing the `CreateRelease` and `UpdateRelease` fixed this

NGitLab/Impl/TagClient.cs
Bromvlieg: I've removed them 😄

NGitLab/IRepositoryClient.cs
```diff
@@ -8,6 +8,7 @@ namespace NGitLab
     public interface IRepositoryClient
     {
         ITagClient Tags { get; }
+        IReleaseClient Releases { get; }
```
jairbubbles: Can you add the new property at the end of the interface?

NGitLab/Impl/ReleaseClient.cs
jairbubbles: Well, you shouldn't feel bad to do better that other classes ;-)

jairbubbles (commented):
Awesome, I just made two minor remarks and we're good to go! 🚀

jairbubbles:
Don't know if you noticed but you have some styling issues:
```
Error: /home/runner/work/NGitLab/NGitLab/NGitLab/IReleaseClient.cs(9,1): error SA1516: Elements should be separated by blank line [/home/runner/work/NGitLab/NGitLab/NGitLab/NGitLab.csproj]
Error: /home/runner/work/NGitLab/NGitLab/NGitLab/IRepositoryClient.cs(11,1): error SA1516: Elements should be separated by blank line [/home/runner/work/NGitLab/NGitLab/NGitLab/NGitLab.csproj]
Error: /home/runner/work/NGitLab/NGitLab/NGitLab/Impl/RepositoryClient.cs(26,1): error SA1516: Elements should be separated by blank line [/home/runner/work/NGitLab/NGitLab/NGitLab/NGitLab.csproj]
Error: /home/runner/work/NGitLab/NGitLab/NGitLab/IReleaseClient.cs(9,1): error SA1516: Elements should be separated by blank line [/home/runner/work/NGitLab/NGitLab/NGitLab/NGitLab.csproj]
Error: /home/runner/work/NGitLab/NGitLab/NGitLab/IRepositoryClient.cs(11,1): error SA1516: Elements should be separated by blank line [/home/runner/work/NGitLab/NGitLab/NGitLab/NGitLab.csproj]
Error: /home/runner/work/NGitLab/NGitLab/NGitLab/Impl/RepositoryClient.cs(26,1): error SA1516: Elements should be separated by blank line [/home/runner/work/NGitLab/NGitLab/NGitLab/NGitLab.csproj]
```

Bromvlieg:
It's a bit odd, my visual studio only complains about it when I have the specific file open, so most of the times while building I don't see the style warnings 🤔

NGitLab/Impl/ReleaseClient.cs
Bromvlieg: Fair enough, added 😸

NGitLab/IRepositoryClient.cs
Bromvlieg: I put it at the first best place where I saw another `Client`, I've moved it down now.

jairbubbles (approved):
Fine with me, we'll just wait for the PR to be green.

jairbubbles:
@Bromvlieg  The tests need to be updated too:
```
Error: /home/runner/work/NGitLab/NGitLab/NGitLab.Tests/TagsTests.cs(47,38): error CS1061: 'ITagClient' does not contain a definition for 'CreateRelease' and no accessible extension method 'CreateRelease' accepting a first argument of type 'ITagClient' could be found (are you missing a using directive or an assembly reference?) [/home/runner/work/NGitLab/NGitLab/NGitLab.Tests/NGitLab.Tests.csproj]
Error: /home/runner/work/NGitLab/NGitLab/NGitLab.Tests/TagsTests.cs(51,34): error CS1061: 'ITagClient' does not contain a definition for 'UpdateRelease' and no accessible extension method 'UpdateRelease' accepting a first argument of type 'ITagClient' could be found (are you missing a using directive or an assembly reference?) [/home/runner/work/NGitLab/NGitLab/NGitLab.Tests/NGitLab.Tests.csproj]
```

jairbubbles:
Thanks a lot @Bromvlieg  for this contribution, you're our first external contribution on this repository and we're super happy.

We would also be glaf to get your feedbacks regarding the contribution flow to see what we could improve!

Bromvlieg:
No problem! Google gave this repository in the results, tried it on NuGet and seemed to work clean and easy. You guys kept the code nice and simple to figure out so I figured why not make a PR to add the releases rather than yolo my own solution with curl.

As feedback perhaps some more samples on how to use the library to go trough all branches to get a quick feel for how it works and what logic/style you can expect when using it.

jairbubbles:
(published in version [6.9.0](https://www.nuget.org/packages/NGitLab/6.9.0))

meziantou:
> It's a bit odd, my visual studio only complains about it when I have the specific file open, so most of the times while building I don't see the style warnings 🤔

You need to change your VS settings to enable solution wide analysis (this uses more resources)

image

PR #40, opened 2021-07-07
Additional functionality for ReleaseClient + ReleaseLinksClient for release links API

dianaMihaelaDiaconu:
Changes to ReleaseClient:
Added all the fields in Release.
The name of the release ca be different than the tag name.
A release can be created with asset links.

Added LinksClient to implement release links API.

Added back CreateRelease and UpdateRelease methods in TagClient. Removing them was a breaking change. (I don't know if there's much sense in adding them back though.)

Added some tests for the new API.

There might be a few changes where there is not much added value (you might wander "What's the point in changing that?") the reason is that I have implemented this before seeing the other commit. I was just about to create a pull request when the other one got merged.

As it is, my code contains some breaking changes. There sure is a way to work around that, but as I mentioned, the previous commit already had some hard breaking changes (the functions that were removed had not been previously deprecated), so first I wanted your input before starting to make changes. Here they are:
- I've chosen to put GetReleases method in GitLabClient, instead of RepositoryClient. For me it made more sense to be there, 
as the api is not related to repository. 
- RelaseClient methods return an object of type "Release", instead of "ReleaseInfo". I wanted to make a distinction between the release info that is returned from the tags api (that is more limited), and the one from the releases. If the functions in tags remain removed, there's no need for the old ReleaseInfo, so I have no problem renaming the new one.
- I created ReleaseCollection class to deal with releases for the mock releaseClient. I didn't want to pollute the repository class, plus it's uniform with how we handle other collections (badges, merge requests, pipelines, etc). But since the logic was already added to Repository class, I guess it's now a breaking change.

### DIFF
--- a/NGitLab.Mock/Clients/ReleaseClient.cs
+++ b/NGitLab.Mock/Clients/ReleaseClient.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using NGitLab.Models;
 
@@ -14,43 +15,88 @@ namespace NGitLab.Mock.Clients
             _projectId = projectId;
         }
 
-        public IEnumerable<ReleaseInfo> All
+        public IEnumerable<Models.ReleaseInfo> All
         {
             get
             {
                 using (Context.BeginOperationScope())
                 {
-                    var project = GetProject(_projectId, ProjectPermission.Contribute);
-                    return project.Repository.GetReleases();
+                    var project = GetProject(_projectId, ProjectPermission.View);
+                    return project.Releases.Select(r => r.ToReleaseClient());
                 }
             }
         }
 
-        public ReleaseInfo Create(ReleaseCreate data)
+        public Models.ReleaseInfo this[string tagName]
         {
-            using (Context.BeginOperationScope())
+            get
             {
-                var project = GetProject(_projectId, ProjectPermission.Contribute);
-                return project.Repository.CreateRelease(data.Name, data.Description);
+                using (Context.BeginOperationScope())
+                {
+                    var project = GetProject(_projectId, ProjectPermission.View);
+                    var release = project.Releases.FirstOrDefault(r => r.TagName.Equals(tagName));
+
+                    return release.ToReleaseClient();
+                }
             }
         }
 
-        public void Delete(string name)
+        public Models.ReleaseInfo Create(Models.ReleaseCreate data)
         {
             using (Context.BeginOperationScope())
             {
-                var project = GetProject(_projectId, ProjectPermission.Contribute);
-                project.Repository.DeleteRelease(name);
+                var project = GetProject(_projectId, ProjectPermission.View);
+                var release = project.Releases.Add(data.TagName, data.Name, data.Description, Context.User);
+                return release.ToReleaseClient();
             }
         }
 
-        public ReleaseInfo Update(ReleaseUpdate data)
+        public Models.ReleaseInfo Update(ReleaseUpdate data)
         {
             using (Context.BeginOperationScope())
             {
-                var project = GetProject(_projectId, ProjectPermission.Contribute);
-                return project.Repository.UpdateRelease(data.Name, data.Description);
+                var project = GetProject(_projectId, ProjectPermission.View);
+                var release = project.Releases.GetByTagName(data.TagName);
+                if (release == null)
+                {
+                    throw new GitLabNotFoundException();
+                }
+
+                if (data.Name != null)
+                {
+                    release.Name = data.Name;
+                }
+
+                if (data.Description != null)
+                {
+                    release.Description = data.Description;
+                }
+
+                if (data.ReleasedAt.HasValue)
+                {
+                    release.ReleasedAt = data.ReleasedAt.Value;
+                }
+
+                return release.ToReleaseClient();
             }
+        }
+
+        public void Delete(string tagName)
+        {
+            using (Context.BeginOperationScope())
+            {
+                var project = GetProject(_projectId, ProjectPermission.View);
+                var release = project.Releases.FirstOrDefault(r => r.TagName.Equals(tagName));
+                if (release == null)
+                    throw new GitLabNotFoundException();
+
+                project.Releases.Remove(release);
+            }
+        }
+
+        public IReleaseLinkClient ReleaseLinks(string tagName)
+        {
+            throw new NotImplementedException();
         }
     }
 }
--- a/NGitLab.Mock/Clients/ReleaseLinkClient.cs
+++ b/NGitLab.Mock/Clients/ReleaseLinkClient.cs
@@ -35,7 +35,5 @@ namespace NGitLab.Mock.Clients
         {
             throw new NotImplementedException();
         }
-
-       // private MergeRequest GetRelease() => GetMergeRequest(_projectId, _mergeRequestIid);
     }
 }
--- a/NGitLab.Mock/Clients/ReleaseLinkClient.cs
+++ b/NGitLab.Mock/Clients/ReleaseLinkClient.cs
@@ -1,0 +1,41 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using NGitLab.Models;
+
+namespace NGitLab.Mock.Clients
+{
+    internal sealed class ReleaseLinkClient : ClientBase, IReleaseLinkClient
+    {
+        private readonly int _projectId;
+        private readonly string _tagName;
+
+        public ReleaseLinkClient(ClientContext context, int projectId, string tagName)
+            : base(context)
+        {
+            _projectId = projectId;
+            _tagName = tagName;
+        }
+
+        public ReleaseLink this[int id] => throw new NotImplementedException();
+
+        public IEnumerable<ReleaseLink> All => throw new NotImplementedException();
+
+        public ReleaseLink Create(ReleaseLinkCreate data)
+        {
+            throw new NotImplementedException();
+        }
+
+        public void Delete(int id)
+        {
+            throw new NotImplementedException();
+        }
+
+        public ReleaseLink Update(int id, ReleaseLinkUpdate data)
+        {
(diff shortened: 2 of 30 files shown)


NGitLab.Mock/Clients/TagClient.cs
```diff
@@ -36,13 +37,14 @@ public Tag Create(TagCreate tag)
...
-        public ReleaseInfo CreateRelease(string name, ReleaseCreate data)
+        [Obsolete("Starting in Gitlab 14 releases cannot be created trough tags. Use `ReleaseClient.Create` instead", false)]
+        public RealeaseInfo CreateRelease(string name, ReleaseCreate data)
         {
             using (Context.BeginOperationScope())
             {
                 var project = GetProject(_projectId, ProjectPermission.Contribute);
                 var tag = project.Repository.CreateReleaseTag(name, data.Description);
-                return new ReleaseInfo
+                return new RealeaseInfo
```
jairbubbles: Please make sure to merge your branche with main, there were recent changes on the subject.

EDIT: I didn' read the description properly, you are aware of the other commit. I feel like we should keep the new class withtout the typo.

jairbubbles:
> Added back CreateRelease and UpdateRelease methods in TagClient. Removing them was a breaking change. (I don't know if there's much sense in adding them back though.)

I don't think they were used so I was ok to remove them.

> RelaseClient methods return an object of type "Release", instead of "ReleaseInfo". I wanted to make a distinction between the release info that is returned from the tags api (that is more limited), and the one from the releases. If the functions in tags remain removed, there's no need for the old ReleaseInfo, so I have no problem renaming the new one.

This is nice but as the old api was removed, it might be simple just to keep `ReleaseInfo`.

dianaMihaelaDiaconu:
> > Added back CreateRelease and UpdateRelease methods in TagClient. Removing them was a breaking change. (I don't know if there's much sense in adding them back though.)
> 
> I don't think they were used so I was ok to remove them.

We were actually using the CreateRelease function. It's how I noticed they were removed. I can replace the usage with the new API, I was just wondering if it was an oversight or intended, as they were not deprecated. Will remove them from this PR.
> 
> > RelaseClient methods return an object of type "Release", instead of "ReleaseInfo". I wanted to make a distinction between the release info that is returned from the tags api (that is more limited), and the one from the releases. If the functions in tags remain removed, there's no need for the old ReleaseInfo, so I have no problem renaming the new one.
> 
> This is nice but as the old api was removed, it might be simple just to keep `ReleaseInfo`.

Will do.

NGitLab.Mock/Clients/TagClient.cs
dianaMihaelaDiaconu: Done, Using ReleaseInfo everywhere

NGitLab/Impl/ReleaseClient.cs
```diff
@@ -4,34 +4,30 @@
 
 namespace NGitLab.Impl
 {
-    // Handles interacting with Release Tags.
-    // Documentation: https://docs.gitlab.com/ee/api/releases/
```
jairbubbles, line 8: Please keep the comment.

dianaMihaelaDiaconu: Done. Also added for ReleaseLinkClient

NGitLab/Models/ReleaseLinkCreate.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Runtime.Serialization;
+
+namespace NGitLab.Models
+{
+    [DataContract]
+    public class ReleaseLinkCreate
+    {
+        [DataMember(Name = "name")]
+        public string Name;
```
meziantou: The project is pretty old and the original code didn't use properties. That's why most classes have public fields. But we do prefer using properties for new classes. You could use auto properties such as:

```suggestion
        public string Name { get; set; }
```

NGitLab/Impl/ReleaseLinkClient.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Collections.Generic;
+using System.Globalization;
+using System.Net;
+using NGitLab.Models;
+
+namespace NGitLab.Impl
+{
+    // Handles interacting with Release links.
+    // Documentation: https://docs.gitlab.com/ee/api/releases/links
+    public class ReleaseLinkClient : IReleaseLinkClient
```
meziantou: The class doesn't need to be public as we only expose the interface

```suggestion
    internal sealed class ReleaseLinkClient : IReleaseLinkClient
```

NGitLab/Impl/ReleaseClient.cs
```diff
@@ -4,34 +4,32 @@
...
+            _projectId = projectId;
+            var projectPath = Project.Url + "/" + projectId;
+            _releasesPath = projectPath + "/releases";
         }
 
-        public ReleaseInfo Create(ReleaseCreate release)
-        {
-            return _api.Post().With(release).To<ReleaseInfo>(_releasePath);
-        }
+        public IEnumerable<ReleaseInfo> All  => _api.Get().GetAll<ReleaseInfo>(_releasesPath + "?per_page=50");
```
meziantou: In other clients we use the default page size for the `All` property. To set the page size, we add another method such as `Get(ReleaseQuery query)`. The query class can expose a `PerPage` property. If we go with the `Get` method, I'm not sure we need the `All` property

NGitLab.Mock/Clients/ReleaseLinkClient.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            throw new NotImplementedException();
+        }
+
+        public ReleaseLink Update(int id, ReleaseLinkUpdate data)
+        {
+            throw new NotImplementedException();
+        }
+
+       // private MergeRequest GetRelease() => GetMergeRequest(_projectId, _mergeRequestIid);
```
meziantou: I think you can delete this comment

meziantou (commented):
Thank you!
I've few minor comments before merging 😊

NGitLab/Models/ReleaseLinkCreate.cs
dianaMihaelaDiaconu: Applied to all the new classes

NGitLab/Impl/ReleaseLinkClient.cs
dianaMihaelaDiaconu: Changed to interal, and did the same for ReleaseClient

NGitLab/Impl/ReleaseClient.cs
dianaMihaelaDiaconu: Turns out the class I picked to mimic was one of the few that has the per page number hardcoded. Removed to use the default

NGitLab.Mock/Clients/ReleaseLinkClient.cs
dianaMihaelaDiaconu: removed

NGitLab.Mock/Clients/ReleaseClient.cs
```diff
@@ -14,43 +15,88 @@ public ReleaseClient(ClientContext context, int projectId)
...
 
-        public ReleaseInfo Create(ReleaseCreate data)
+        public Models.ReleaseInfo this[string tagName]
+        {
+            get
+            {
+                using (Context.BeginOperationScope())
+                {
+                    var project = GetProject(_projectId, ProjectPermission.View);
+                    var release = project.Releases.FirstOrDefault(r => r.TagName.Equals(tagName));
```
louis-z, line 37:
```suggestion
                    var release = project.Releases.FirstOrDefault(r => r.TagName.Equals(tagName, StringComparison.Ordinal));
```

NGitLab.Mock/ReleaseCollection.cs
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        public override void Add(ReleaseInfo release)
+        {
+            if (release is null)
+                throw new ArgumentNullException(nameof(release));
+
+            if (release.TagName == default || Project.Repository.GetTags().FirstOrDefault(t => t.FriendlyName.Equals(release.TagName)) == null)
+            {
+                throw new ArgumentException(nameof(release));
```
louis-z, line 27: Please provide a message as first argument, something like `$"Tag '{release.TagName}' invalid or not found"`.

NGitLab.Mock/Clients/ReleaseClient.cs
```diff
@@ -14,43 +15,88 @@ public ReleaseClient(ClientContext context, int projectId)
...
 
-        public ReleaseInfo Update(ReleaseUpdate data)
+        public void Delete(string tagName)
         {
             using (Context.BeginOperationScope())
             {
-                var project = GetProject(_projectId, ProjectPermission.Contribute);
-                return project.Repository.UpdateRelease(data.Name, data.Description);
+                var project = GetProject(_projectId, ProjectPermission.View);
+                var release = project.Releases.FirstOrDefault(r => r.TagName.Equals(tagName));
```
louis-z, line 89:
```suggestion
                var release = project.Releases.FirstOrDefault(r => r.TagName.Equals(tagName, StringComparison.Ordinal));
```

NGitLab.Mock/ReleaseCollection.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            return this.FirstOrDefault(r => r.TagName.Equals(tagName));
+        }
+
+        public override void Add(ReleaseInfo release)
+        {
+            if (release is null)
+                throw new ArgumentNullException(nameof(release));
+
+            if (release.TagName == default || Project.Repository.GetTags().FirstOrDefault(t => t.FriendlyName.Equals(release.TagName)) == null)
```
louis-z, line 25:
```suggestion
            if (release.TagName == default || Project.Repository.GetTags().FirstOrDefault(t => t.FriendlyName.Equals(release.TagName, StringComparison.Ordinal)) == null)
```

louis-z (changes requested):

PR #41, opened 2021-07-07
Add labels api support for mocks

LePtitDev:


NGitLab.Mock.Tests/LabelsMockTests.cs
```diff
@@ -0,0 +1,143 @@
...
+using System.Linq;
+using NGitLab.Models;
+using NUnit.Framework;
+
+namespace NGitLab.Mock.Tests
+{
+    public class LabelsMockTests
+    {
+        [Test]
+        public void Test_labels_from_project_can_be_found()
```
meziantou: note for later: we should really find a way to run the same tests for the server and the mock. This would increase the confidence in the mocks

NGitLab.Mock/Clients/LabelClient.cs
```diff
@@ -11,49 +12,109 @@ public LabelClient(ClientContext context)
         {
         }
 
-        public Label Create(LabelCreate label)
+        public Models.Label Create(LabelCreate label)
         {
-            throw new NotImplementedException();
+            var project = FindProject(label.Id) ?? throw new InvalidOperationException($"Cannot find project with ID {label.Id}");
```
meziantou: You need to wrap method calls into

````
````

File: NGitLab.Mock/Clients/LabelClient.cs [hunk 3]
```diff
@@ -11,49 +12,109 @@ public LabelClient(ClientContext context)
...
+            return project.Labels.Add(label.Name, label.Color, label.Description).ToClientLabel();
         }
 
-        public Label CreateGroupLabel(LabelCreate label)
+        public Models.Label CreateGroupLabel(LabelCreate label)
         {
-            throw new NotImplementedException();
+            var group = FindGroup(label.Id) ?? throw new InvalidOperationException($"Cannot find group with ID {label.Id}");
```
meziantou: You must throw specific GitLab exceptions such as NotFound. Also you need to check for permissions. (you can check how GetProject is implemented)

NGitLab.Mock/Clients/LabelClient.cs
```diff
@@ -11,49 +12,109 @@ public LabelClient(ClientContext context)
...
-        public IEnumerable<Label> ForGroup(int groupId)
+        public Models.Label GetLabel(int projectId, string name)
         {
-            throw new NotImplementedException();
+            var project = FindProject(projectId) ?? throw new InvalidOperationException($"Cannot find project with ID {projectId}");
+            return FindLabel(project.Labels, name)?.ToClientLabel();
         }
 
-        public IEnumerable<Label> ForProject(int projectId)
+        private Project FindProject(int id)
```
meziantou: Not needed, you should be able to use the `GetProject` method defined in the base class

meziantou (changes requested):

NGitLab.Mock.Tests/LabelsMockTests.cs
jairbubbles: Indeed, it would be awesome!

NGitLab.Mock/LabelsCollection.cs
```diff
@@ -0,0 +1,58 @@
...
+        public override void Add(Label label)
+        {
+            if (label is null)
+                throw new ArgumentNullException(nameof(label));
+
+            if (label.Id == default)
+            {
+                label.Id = Server.GetNewLabelId();
+            }
+            else if (GetById(label.Id) != null)
```
meziantou, line 55: Are labels unique by name?

LePtitDev: I don't know, it's not specified in the [doc](https://docs.gitlab.com/ee/api/labels.html#create-a-new-label).

meziantou: I've tested using the web interface and you get an error : `Title has already been taken`

meziantou (approved):

NGitLab.Mock.Tests/LabelsMockTests.cs
```diff
@@ -0,0 +1,143 @@
...
+
+            var client = gitLabServer.CreateClient(user1);
+            client.Labels.Delete(new LabelDelete { Id = project.Id, Name = "test1" });
+            var labels = client.Labels.ForProject(project.Id).ToArray();
+
+            Assert.AreEqual(0, labels.Length, "Labels count is invalid");
+        }
+
+        [Test]
+        public void Test_labels_from_group_can_be_found()
```
louis-z:
```suggestion
        public void Test_labels_can_be_found_from_group()
```

NGitLab.Mock.Tests/LabelsMockTests.cs
louis-z:
```suggestion
        public void Test_labels_can_be_found_from_project()
```

louis-z (changes requested):

PR #46, opened 2021-07-09
Modify RepositoryMock's behavior to fit GitLab's

mogawinch:
Based on this: https://docs.gitlab.com/ee/api/releases/#create-a-release

When a tag with 'tag_name' does not exist and 'ref' exist, a tag with 'tag_name' will be created when the release will be created.


NGitLab.Mock/ReleaseCollection.cs
```diff
@@ -40,13 +41,21 @@ public override void Add(ReleaseInfo release)
             base.Add(release);
         }
 
-        public ReleaseInfo Add(string tagName, string name, string description, User user)
+        public ReleaseInfo Add(string tagName, string name, string reference, string description, User user)
```
louis-z, line 49: This is a breaking change. Should we add a method overload instead?

@meziantou , what do you think?

NGitLab.Mock/ReleaseCollection.cs
```diff
@@ -40,13 +41,21 @@ public override void Add(ReleaseInfo release)
...
         {
             if (tagName is null)
                 throw new ArgumentNullException(nameof(tagName));
             if (user is null)
                 throw new ArgumentNullException(nameof(user));
 
+            if (!Project.Repository.GetTags().Any(r => string.Equals(r.FriendlyName, tagName, StringComparison.Ordinal)))
+            {
+                if (string.IsNullOrEmpty(reference))
+                    throw new GitLabBadRequestException();
```
louis-z: Please add a message indicating why the exception is thrown.

louis-z (changes requested):

NGitLab.Mock/ReleaseCollection.cs
jairbubbles: (on mocks, I'm not sure we care as long as it's source compatible)

jairbubbles: (would be nice to add https://www.nuget.org/packages/Microsoft.CodeAnalysis.PublicApiAnalyzers)

NGitLab.Mock/ReleaseCollection.cs
meziantou: done

NGitLab.Mock/ReleaseCollection.cs
meziantou: I've reintroduced the previous method, so there is no breaking change.

@jairbubbles https://github.com/ubisoft/NGitLab/issues/52

louis-z (approved):

PR #53, opened 2021-07-14
Add PublicApiAnalyzers

### LINKED ISSUE #52: Use Microsoft.CodeAnalysis.PublicApiAnalyzers to prevent breaking changes
We should add [Microsoft.CodeAnalysis.PublicApiAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.PublicApiAnalyzers) to ensure we do not introduce breaking changes to NGitLab and NGitLab.Mock

meziantou:
Fix #52


meziantou:
@jairbubbles Do you think we should use the "shipped" file. In this project all changes are deployed almost immediately, so all changes should go to the shipped file. But the fixer add new APIs to the unshipped, so it is easier to add them to this file.

louis-z:
> @jairbubbles Do you think we should use the "shipped" file. In this project all changes are deployed almost immediately, so all changes should go to the shipped file. But the fixer add new APIs to the unshipped, so it is easier to add them to this file.

What's the recommended process? To move stuff from Unshipped to Shipped just before tagging?

meziantou:
When you change the public API, new changes should go to the unshipped file. When you release (tag), you should move all changes to shipped.

While we always tag, one of the file seems useless.

jairbubbles:
@meziantou Indeed the Shipped / Unshipped does not make much sense in our case but it's not a big deal.

PR #54, opened 2021-07-19
Log requests and responses in tests

meziantou:


louis-z (commented):
- Is the main intent of this PR to provide us with more info to debug flaky tests?
- Is there a way to validate the new code? Perhaps having a test that sends out a request we know will fail and then asserting data that was logged in the process?

meziantou:
> Is the main intent of this PR to provide us with more info to debug flaky tests?

Yes, it is the goal :)

> Is there a way to validate the new code? Perhaps having a test that sends out a request we know will fail and then asserting data that was logged in the process?

I've tested the code manually by making a test fails. I'm not sure we can test the content of `TestContext.WriteLine`.

PR #57, opened 2021-07-20
Ensure tests only send one request at a time to GitLab

meziantou:


NGitLab.Tests/Docker/GitLabTestContextRequestOptions.cs
```diff
@@ -33,86 +35,102 @@ public override WebResponse GetResponse(HttpWebRequest request)
...
+            s_semaphoreSlim.Wait();
             try
             {
-                response = base.GetResponse(request);
-            }
-            catch (WebException exception)
-            {
-                response = exception.Response;
-                if (response is HttpWebResponse webResponse)
+                try
```
louis-z, line 43: Do we really need to nest the try/catch/finally blocks? Could we not simply release the semaphore from the same finally block where the call to `response = LogRequest(request, response)` is made from?

meziantou: LogRequest is using the response stream, so it could fail. I think it's safer to keep it in an outer try/finally block

louis-z: ...unless of course the said call might actually throw.

louis-z: Oops, forgot to refresh before posting my last comment. 🙄

PR #64, opened 2021-07-28
Added DueDate issue param

apoufortin:
Added the due_date field from Gitlab
DueDate does not contain a time section, so a new time formatter was added.

### DIFF
--- a/NGitLab/Models/Issue.cs
+++ b/NGitLab/Models/Issue.cs
@@ -45,6 +45,9 @@ namespace NGitLab.Models
         [DataMember(Name = "updated_at")]
         public DateTime UpdatedAt;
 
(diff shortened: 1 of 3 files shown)


NGitLab/Models/Issue.cs
```diff
@@ -45,6 +45,9 @@ public class Issue
         [DataMember(Name = "updated_at")]
         public DateTime UpdatedAt;
 
+        [DataMember(Name = "due_date")]
+        public DateTime? DueDate;
```
meziantou, line 49: You need to fix the warning / error in Visual Studio. The code fix should add a new line in `NGitLab/PublicAPI.Unshipped.txt`

PR #65, opened 2021-07-29
Add missing variable properties

Toa741:


NGitLab/Models/Variable.cs
```diff
@@ -13,5 +13,14 @@ public class Variable
 
         [DataMember(Name = "protected")]
         public bool Protected;
+
+        [DataMember(Name = "variable_type")]
+        public VariableType Type;
```
louis-z: This introduces a breaking change. You'll need to add an entry to `NGitLab\PublicAPI.Unshipped.txt`, but this can easily be done by using the automated code fix provided within VS.

Furthermore, since you're already introducing a breaking change, we could take the opportunity to convert all public fields to properties. What do you think?

louis-z (changes requested):

NGitLab/Models/Variable.cs
Toa741: VS Suggestions seems to work fine ^^.

Sure, no problem to make them as properties.

louis-z (approved):

PR #67, opened 2021-08-02
Add config models with fluent methods to simplify mocks creation

LePtitDev:


NGitLab.Mock/Fluent/GitLabUser.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace NGitLab.Mock.Fluent
+{
+    public class GitLabUser : GitLabObject<GitLabConfig>
+    {
+        public string Username { get; set; }
+
+        public string Name { get; set; } = "User";
+
+        public string Email { get; set; } = "user@ubisoft.com";
```
meziantou: You should use "example.com" domain

```suggestion
        public string Email { get; set; } = "user@example.com";
```

meziantou:
Why do we need all these new methods. Can't we just add new overloads to existing objects (`GitLabServer.Users.Add(...)`)?
Can you share an example of use to see the difference between the existing way to configure objects and the fluent way?

meziantou:
@jairbubbles Instead of adding lots of new methods, what do you think of improving existing collections to support a more friendly way to add items:

````c#
using var server = new GitLabServer();
server.Users.AddNew(); // current method
server.Users.AddNew("username"); // current method

server.Users.AddNew(user => { ... }); // proposed method
server.Users.AddNew("username", user => { ... }); // proposed method
// we can use similar methods for other types
````

**Comment by jairbubbles** [comment]
@meziantou Yes it would make sense but my understanding is that the new fluent methods proposed by @LePtitDev  were not only about the "fluent" part, they were also more high level and were simplifying the configuration. I think we should discuss it all together.

File: NGitLab.Mock.Tests/CommitsMockTests.cs [hunk 2]
```diff
@@ -0,0 +1,58 @@
...
+
+        [Test]
+        public void Test_commits_with_tags_can_be_found()
+        {
+            using var server = new GitLabConfig()
+                .WithUser("user1", asDefault: true)
+                .WithProject("test-project", id: 1, configure: project => project
+                    .WithCommit("Initial commit")
+                    .WithCommit("Changes with tag", tags: new[] { "1.0.0" }))
+                .ResolveServer();
```
meziantou:
```suggestion
                .BuildServer();
```

NGitLab.Mock.Tests/CommitsMockTests.cs
```diff
@@ -0,0 +1,58 @@
...
+
+            var client = server.ResolveClient();
```
meziantou:
```suggestion
            var client = server.CreateClient();
```

NGitLab.Mock.Tests/MilestonesMockTests.cs
```diff
@@ -73,49 +62,39 @@ public void Test_milestones_can_be_edited_from_project()
...
-            var milestone = new Milestone { Title = "Milestone 1" };
-            project.Milestones.Add(milestone);
-
-            var client = gitLabServer.CreateClient(user1);
-            client.GetMilestone(project.Id).Close(milestone.Id);
-            var activeMilestones = client.GetMilestone(project.Id).AllInState(Models.MilestoneState.active).ToArray();
-            var closedMilestones = client.GetMilestone(project.Id).AllInState(Models.MilestoneState.closed).ToArray();
+            using var server = new GitLabConfig()
+                .WithUser("user1", asDefault: true)
+                .WithProject("Test", id: 1, defaultAsMaintainer: true, configure: project => project
```
meziantou: The project name should be optional (randomly generated if not set), so you don't need to set it when it doesn't impact the test

NGitLab.Mock/Config/GitLabHelpers.cs
```diff
@@ -0,0 +1,1279 @@
...
+        {
+            return WithMilestone(project, title, milestone =>
+            {
+                if (id != default)
+                    milestone.Id = id;
+
+                milestone.Description = description;
+                milestone.DueDate = dueDate;
+                milestone.StartDate = startDate;
+                milestone.CreatedAt = createdAt ?? DateTime.Now;
```
meziantou: Add dates should be Utc

```suggestion
                milestone.CreatedAt = createdAt ?? DateTime.UtcNow;
```

NGitLab.Mock/Config/GitLabLabel.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// Describe a label in a GitLab group/project
+    /// </summary>
+    public class GitLabLabel : GitLabObject
+    {
+        public string Name { get; set; }
+
+        /// <summary>
+        /// Color in RGB hex format (example: #5884AD)
+        /// </summary>
+        public string Color { get; set; } = "#d9534f";
```
meziantou: Why do we need a default value?

NGitLab.Mock/Config/GitLabMergeRequest.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// Labels names
+        /// </summary>
+        public IList<string> Labels { get; } = new List<string>();
+
+        /// <summary>
+        /// Approvers usernames
+        /// </summary>
+        public IList<string> Approvers { get; } = new List<string>();
+
+        public DateTime CreatedAt { get; set; } = DateTime.Now;
```
meziantou:
```suggestion
        public DateTime CreatedAt { get; set; } = DateTime.UtcNow;
```

NGitLab.Mock/Config/GitLabObject.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace NGitLab.Mock.Config
+{
+    public abstract class GitLabObject
+    {
+        protected internal GitLabObject()
+        {
+        }
+
+        public int Id { get; set; }
```
meziantou, line 11: Not all GitLab objects have a numeric id. For instance templates, deployments, experiments, feature flags, etc.

NGitLab.Mock/Config/GitLabPermission.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// Username (required if user permission)
+        /// </summary>
+        public string User { get; set; }
+
+        /// <summary>
+        /// Group fullname (required is group permission)
+        /// </summary>
+        public string Group { get; set; }
+
+        public AccessLevel Level { get; set; } = AccessLevel.Developer;
```
meziantou: The value is already set by the extension methods

```suggestion
        public AccessLevel Level { get; set; }
```

NGitLab.Mock/Config/GitLabProject.cs
```diff
@@ -0,0 +1,58 @@
...
+        public string DefaultBranch { get; set; } = "main";
+
+        public string Description { get; set; }
+
+        /// <summary>
+        /// Path where to clone repository after server resolving
+        /// </summary>
+        public string ClonePath { get; set; }
+
+        public VisibilityLevel Visibility { get; set; } = VisibilityLevel.Internal;
```
meziantou: The default visibility is configured on the server. So, this property should read the server value.

NGitLab.Mock/readme.md
```diff
@@ -7,6 +7,30 @@ It allows to write simple unit tests on code that uses GitLab.
 
 # How to use it?
 
+## 1. With config (*RECOMMENDED*)
```
meziantou, line 10: Can you add a line to explain why this is recommended?

NGitLab.Mock/Config/ConfigSerializer.cs
```diff
@@ -0,0 +1,426 @@
...
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using YamlDotNet.Serialization;
+
+#pragma warning disable MA0009 // Add regex evaluation timeout
+
+namespace NGitLab.Mock.Config
+{
+    internal class ConfigSerializer
```
meziantou, line 14: Isn't YamlDotNet able to serialize the objects without all of this code?

NGitLab.Mock/Config/GitLabConfig.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            Users = new GitLabUsersCollection(this);
+            Groups = new GitLabGroupsCollection(this);
+            Projects = new GitLabProjectsCollection(this);
+        }
+
+        /// <summary>
+        /// User used by default when not specified in extensions methods
+        /// </summary>
+        public string DefaultUser { get; set; }
```
meziantou, line 20: Should setters (almost in every object) be internal and only used by the extension methods?

NGitLab.Mock/Config/GitLabHelpers.cs
```diff
@@ -0,0 +1,1279 @@
...
+        /// </summary>
+        /// <param name="config">Config.</param>
+        /// <param name="username">Username (required)</param>
+        /// <param name="name">Name.</param>
+        /// <param name="email">Email.</param>
+        /// <param name="avatarUrl">Avatar URL.</param>
+        /// <param name="isAdmin">Indicates if user is admin in GitLab server</param>
+        /// <param name="asDefault">Define as default user for next descriptions</param>
+        /// <param name="configure">Configuration method</param>
+        public static GitLabConfig WithUser(this GitLabConfig config, string username, string name = "User", string email = "user@example.com", string avatarUrl = null, bool isAdmin = false, bool asDefault = false, Action<GitLabUser> configure = null)
```
meziantou:
```suggestion
        public static GitLabConfig WithUser(this GitLabConfig config, string username, string name = "User", string email = "user@example.com", string avatarUrl = null, bool isAdmin = false, bool isDefault = false, Action<GitLabUser> configure = null)
```

NGitLab.Mock/Config/GitLabHelpers.cs
```diff
@@ -0,0 +1,1279 @@
...
+        /// </summary>
+        /// <param name="config">Config.</param>
+        /// <param name="name">Name (required)</param>
+        /// <param name="id">Explicit ID (config increment)</param>
+        /// <param name="namespace">Parent namespace.</param>
+        /// <param name="description">Description.</param>
+        /// <param name="visibility">Visibility.</param>
+        /// <param name="defaultAsMaintainer">Define default user as maintainer.</param>
+        /// <param name="configure">Configuration method</param>
+        public static GitLabConfig WithGroup(this GitLabConfig config, string name, int id = default, string @namespace = null, string description = null, VisibilityLevel visibility = VisibilityLevel.Internal, bool defaultAsMaintainer = false, Action<GitLabGroup> configure = null)
```
meziantou: I think `defaultAsMaintainer` isn't clear enough. I had to read the code to understand what is does.

```suggestion
        public static GitLabConfig WithGroup(this GitLabConfig config, string name, int id = default, string @namespace = null, string description = null, VisibilityLevel visibility = VisibilityLevel.Internal, bool addDefaultUserAsMaintainer = false, Action<GitLabGroup> configure = null)
```

meziantou:
```
VisibilityLevel visibility = VisibilityLevel.Internal
```

I think the default visibility should be set on the server (as GitLab), and reused here. Maybe it needs a new enum value `Default` (and a new enum)

meziantou: Shouldn't we have another extension to create subgroups like:

````
public static GitLabGroup WithGroup(this GitLabGroup @parentGroup, string name, ...)
````

File: NGitLab.Mock/Config/GitLabHelpers.cs [hunk 16]
```diff
@@ -0,0 +1,1279 @@
...
+            });
+        }
+
+        /// <summary>
+        /// Add a project description in config
+        /// </summary>
+        /// <param name="config">Config.</param>
+        /// <param name="name">Name (required)</param>
+        /// <param name="configure">Configuration method</param>
+        public static GitLabConfig WithProject(this GitLabConfig config, string name, Action<GitLabProject> configure)
```
meziantou: Is the name really mandatory? What if you are only interested in having an id in your tests. Maybe we should add another method:

````
public static GitLabConfig WithProject(this GitLabConfig config, int id, Action<GitLabProject> configure)
````

File: NGitLab.Mock/Config/GitLabHelpers.cs [hunk 17]
```diff
@@ -0,0 +1,1279 @@
...
+        /// <param name="id">Explicit ID (config increment)</param>
+        /// <param name="namespace">Group fullname.</param>
+        /// <param name="description">Description.</param>
+        /// <param name="defaultBranch">Repository default branch.</param>
+        /// <param name="visibility">Visibility.</param>
+        /// <param name="initialCommit">Indicates if an initial commit is added.</param>
+        /// <param name="defaultAsMaintainer">Define default user as maintainer.</param>
+        /// <param name="clonePath">Path where to clone repository after server resolving</param>
+        /// <param name="configure">Configuration method</param>
+        public static GitLabConfig WithProject(this GitLabConfig config, string name, int id = default, string @namespace = "functional", string description = null, string defaultBranch = "main", VisibilityLevel visibility = VisibilityLevel.Internal, bool initialCommit = false, bool defaultAsMaintainer = false, string clonePath = null, Action<GitLabProject> configure = null)
```
meziantou:
```suggestion
        public static GitLabConfig WithProject(this GitLabConfig config, string name, int id = default, string @namespace = "functional", string description = null, string defaultBranch = "main", VisibilityLevel visibility = VisibilityLevel.Internal, bool initialCommit = false, bool addDefaultUserAsMaintainer = false, string clonePath = null, Action<GitLabProject> configure = null)
```

NGitLab.Mock/Config/GitLabHelpers.cs
```diff
@@ -0,0 +1,1279 @@
...
+            });
+        }
+
+        /// <summary>
+        /// Add group permission in group
+        /// </summary>
+        /// <param name="grp">Group.</param>
+        /// <param name="group">Group fullname (required)</param>
+        /// <param name="level">Access level.</param>
+        public static GitLabGroup WithGroupPermission(this GitLabGroup grp, string group, AccessLevel level = AccessLevel.Developer)
```
meziantou: Is it groupName or groupId or both are accepted? I think the parameter name should be improved

NGitLab.Mock/Config/GitLabHelpers.cs
```diff
@@ -0,0 +1,1279 @@
...
+            });
+        }
+
+        /// <summary>
+        /// Add user permission in group
+        /// </summary>
+        /// <param name="group">Group.</param>
+        /// <param name="user">Username (required)</param>
+        /// <param name="level">Access level.</param>
+        public static GitLabGroup WithUserPermission(this GitLabGroup group, string user, AccessLevel level = AccessLevel.Developer)
```
meziantou: I don't think `level` should be optional

NGitLab.Mock/Config/GitLabIssue.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        /// <summary>
+        /// Author username (required)
+        /// </summary>
+        public string Author { get; set; }
+
+        /// <summary>
+        /// Assignee username
+        /// </summary>
+        public string Assignee { get; set; }
```
meziantou, line 31: An issue can have multiple assignees

NGitLab.Mock/Config/GitLabIssue.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        /// <summary>
+        /// Milestone title
+        /// </summary>
+        public string Milestone { get; set; }
+
+        public DateTime CreatedAt { get; set; } = DateTime.Now;
```
meziantou:
```suggestion
        public DateTime CreatedAt { get; set; } = DateTime.UtcNow;
```

meziantou: BTW, I don't know if you need to set them, or just let the GitLabServer set the values when building the server.

NGitLab.Mock/Config/GitLabMergeRequest.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        /// <summary>
+        /// Author username (required)
+        /// </summary>
+        public string Author { get; set; }
+
+        /// <summary>
+        /// Assignee username
+        /// </summary>
+        public string Assignee { get; set; }
```
meziantou, line 26: An merge request can have multiple assignees

NGitLab.Mock/Config/GitLabProject.cs
```diff
@@ -0,0 +1,58 @@
...
+            Labels = new GitLabLabelsCollection(this);
+            Milestones = new GitLabMilestonesCollection(this);
+        }
+
+        /// <summary>
+        /// Name (required)
+        /// </summary>
+        public string Name { get; set; }
+
+        public string Namespace { get; set; } = "functional";
```
meziantou: You have _constants_ reused across multiple files. You should extract them as actual constants

NGitLab.Mock/Config/GitLabProject.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        public string DefaultBranch { get; set; } = "main";
```
meziantou: I think the server will set the default branch if not set. Do we need to set a default value here?

NGitLab.Mock/Repository.cs
```diff
@@ -355,7 +355,7 @@ public Commit GetCommit(string reference)
 
             var tag = repository.Tags[reference];
             if (tag?.PeeledTarget is Commit commit)
-                return branchTip;
+                return commit;
```
meziantou: Feel free to create another MR to fix this bug

NGitLab.Mock/Config/GitLabMergeRequest.cs
LePtitDev: NGitLab does not implement multiple assignees for merge requests, only for issues

NGitLab.Mock/Config/GitLabProject.cs
LePtitDev: It's because when we deserialize theses properties must be set but optional in the YAML file

NGitLab.Mock/Config/GitLabObject.cs
LePtitDev: All objects that I implemented in config has an ID

NGitLab.Mock/Config/ConfigSerializer.cs
LePtitDev: I extend parsing to support variables combinaisons, environement variables, property name with invalid case, etc.

NGitLab.Mock/Config/GitLabHelpers.cs
LePtitDev: If you set `@namespace` variable, it will resolve parent or create it if not exists

PR #73, opened 2021-08-12
Allow get repository commits in mock

SamuelTrottier:


NGitLab.Mock/Repository.cs
```diff
@@ -323,7 +323,7 @@ public IEnumerable<Commit> GetBranchCommits(string branchName)
             }
         }
 
-        public IEnumerable<Commit> GetCommits(string @ref)
+        public IEnumerable<Commit> GetCommits(string @ref = "")
```
meziantou: I think this is a breaking change (the CI will confirm :)). You need to create a new overload without the parameter. Also, prefer `null` to `""`

NGitLab.Mock/Clients/RepositoryClient.cs
```diff
@@ -87,5 +97,9 @@ public IEnumerable<Ref> GetCommitRefs(Sha1 sha, CommitRefType type = CommitRefTy
         {
             throw new NotImplementedException();
         }
+        private Commit ConvertToNgitlabCommit(LibGit2Sharp.Commit commit, Project project)
```
meziantou:
```suggestion

        private Commit ConvertToNGitLabCommit(LibGit2Sharp.Commit commit, Project project)
```

NGitLab.Mock/Repository.cs
SamuelTrottier: Got it, I was not sure if it was better to add a new overload, i'll add one!

NGitLab.Mock/Repository.cs
```diff
@@ -323,6 +323,11 @@ public IEnumerable<Commit> GetBranchCommits(string branchName)
             }
         }
 
+        public IEnumerable<Commit> GetCommits()
+        {
+            return GetCommits(string.Empty);
```
meziantou, line 328:
```suggestion
            return GetCommits(null);
```

SamuelTrottier: I don't think I can make it null, it would be ambiguous which overload I'm trying to call.

PR #81, opened 2021-08-19
Add Filter to support search without using search=

AlfonsoCampodonico:
Hello, i recently used this nuget package, and noticed that i couldn't use it to filter users like in the UsersAPI pages 
image

i made a small change so it "Filters" instead of searching, so when you call your gitlab API you can filter directly from the URL like in the documentation.

image

https://docs.gitlab.com/ee/api/users.html#for-normal-users


meziantou:
I don't think allowing raw queries is a good thing. I would prefer improving the [`UserQuery` class](https://github.com/ubisoft/NGitLab/blob/main/NGitLab/Models/UserQuery.cs) with the missing properties. This increases the discoverability and prevent errors.

AlfonsoCampodonico:
I think it would be better to use raw queries like the GitLab Docs suggests instead of ?search= 

image

I didn't know about the UserQuery class, looks really neat, if you want it could be extended to use more properties

meziantou:
> I didn't know about the UserQuery class, looks really neat, if you want it could be extended to use more properties

Yep, you can improve the UserQuery class with the missing properties and use the new properties in the [Get method](https://github.com/ubisoft/NGitLab/blob/c4fa8376572f1da4965fc4e9f0f11307b116813e/NGitLab/Impl/UserClient.cs#L26). Then, I don't think there is a need for the `Filter` method :)

AlfonsoCampodonico:
Recently added some more properties, think i missed a couple, will continue tomorrow! Thanks @meziantou !

meziantou:
Good! Let us know when the PR is ready for review or if you need help 😊

AlfonsoCampodonico:
@meziantou  it's ready for review :)

NGitLab/Models/UserQuery.cs
```diff
@@ -46,5 +46,52 @@ public class UserQuery
...
+        /// (Admin only) Search for users without projects
+        /// </summary>
+        public bool? WithoutProjects { get; set; }
+
+
+        /// <summary>
+        /// (Admin only) Search for users created before specified date time
+        /// Format: "YYYY-MM-DDThh:mm:ss.sTZD" eg. "2001-01-02T00:00:00.060Z"
+        /// </summary>
+        public string CreatedBefore { get; set; }
```
meziantou: Should be 

```suggestion
        public DateTime? CreatedBefore { get; set; }
```

NGitLab/Models/UserQuery.cs
```diff
@@ -46,5 +46,52 @@ public class UserQuery
...
+
+        /// <summary>
+        /// (Admin only) Search for users created after specified date time
+        /// Format: "YYYY-MM-DDThh:mm:ss.sTZD" eg. "2001-01-02T00:00:00.060Z"
+        /// </summary>
+        public string CreatedAfter { get; set; }
```
meziantou:
```suggestion
        public DateTime? CreatedAfter { get; set; }
```

NGitLab/Models/UserQuery.cs
```diff
@@ -46,5 +46,52 @@ public class UserQuery
...
+        /// (Admin only) Include the users custom attributes in the response
+        /// </summary>
+        public bool? WithCustomAttributes {  get; set; }
+
+
+        /// <summary>
+        /// (Admin only) Gets users that have two factor either enabled or disabled
+        /// Possible values: "enabled", "disabled"
+        /// </summary>
+        public string TwoFactor { get; set;}
```
meziantou: I think it would be easier to use if we create an enum

```suggestion
        public TwoFactorState? TwoFactor { get; set;}
```

````c#
public enum TwoFactorState
{
    [EnumMember("enabled")]
    Enabled,
    [EnumMember("disabled")]
    Disabled,
}
````

**Review by meziantou — CHANGES_REQUESTED** [review]

**Comment by meziantou** [comment]
Concerning the [errors RS0016 in the CI](https://github.com/ubisoft/NGitLab/pull/81/checks?check_run_id=3382390392#step:4:26), you need to apply the fix for the warning in Visual Studio

**Review by meziantou — APPROVED** [review]

File: NGitLab/Models/User.cs [hunk 4]
```diff
@@ -77,5 +107,48 @@ public class User
         [Obsolete("This does not match GitLab's Api. Use Identities.ExternUid instead.")]
         [DataMember(Name = "extern_uid")]
         public string ExternUid;
+
+        [Obsolete("This does not match GitLab's Api. Use State instead")]
```
meziantou: Why is this member obsolete?

AlfonsoCampodonico: Doesn't appear in JSON, Gitlab API V4 checks if user is blocked with "state"  attribute

AlfonsoCampodonico: Examples:

image
image

AlfonsoCampodonico:
Quick question, why is data member used instead of JsonProperty? Is it because of NET 4.5 Support?

meziantou:
> Quick question, why is data member used instead of JsonProperty? Is it because of NET 4.5 Support?

Because we use SimpleJson. This is an old project, System.Text.Json didn't exist at that time, and I'm pretty sure Newtonsoft was not as good as it is now

NGitLab/Models/User.cs
meziantou: I see. We don't have a clear way to indicate a member exists in specific versions of GitLab. We still have users on 13.x, so adding obsolete is not a great way to do it.
I prefer you remove it from this PR, and we'll think about a better way of obsoleting members later.

NGitLab/Models/UserQuery.cs
meziantou: What do you think about this suggestion?

AlfonsoCampodonico: I think is good, just added it. Did not see it before

NGitLab/Models/User.cs
AlfonsoCampodonico: Made a new commit, removing the obsolete

NGitLab/Models/UserQuery.cs
```diff
@@ -46,5 +49,56 @@ public class UserQuery
...
+        /// Possible values: "enabled", "disabled"
+        /// </summary>
+        public TwoFactorState? TwoFactor { get; set; }
+
+        /// <summary>
+        /// (Admin only)If true, get only users that are Admin
+        /// </summary>
+        public bool? IsAdmin { get; set; }
+
+        public enum TwoFactorState
```
meziantou: You should move the enum out of the `User` type. Nested types are often harder to use/discover.

meziantou:

AlfonsoCampodonico:
If there is anything else i could help with, i would be delighted to

PR #83, opened 2021-09-01
Fix mocked project clone when namespace is not defined

LePtitDev:


meziantou:
@LePtitDev You should not assign the PR manually (except you have a good reason to do so). GitHub automatically assigns 1 reviewer who will manage your PR. This way we ensure the PR are assign to all reviewers using a round robin algorithm.

PR #84, opened 2021-09-02
Allow to delete branch after merge commit in project mock

LePtitDev:


NGitLab.Mock/Config/GitLabCommit.cs
```diff
@@ -30,6 +30,11 @@ public class GitLabCommit : GitLabObject<GitLabProject>
         /// Target branch for a merge commit (required for merge commit)
         /// </summary>
         public string TargetBranch { get; set; }
+
+        /// <summary>
+        /// Indicates if source branch must be deleted after merge
+        /// </summary>
+        public bool DeleteSourceBranch { get; set; }
```
meziantou, line 37: - Should this property be in GitLabMergeRequest?
- I think it should be nullable and also exposed on the Project

    image

LePtitDev: This property is only for Git graph and no related to merge requests. You can construct the project Git graph with config:

```csharp
var config = new GitLabConfig()
    .WithProject("MyProject", configure: project => project
        .WithCommit("Commit on main #1") // Commit 1 on main
        .WithCommit("Commit on branch #2", sourceBranch: "branch") // Commit 2 on branch
        .WithCommit("Commit on main #3") // Commit 3 on main
        .WithMergeCommit("branch", deleteSourceBranch: true)); // Commit 4 merge branch into main
```

PR #86, opened 2021-09-02
Adding expires_at to the add member route

SamuelTrottier:
Adding the expires_at parameter which is supported by the Gitlab API: https://docs.gitlab.com/ee/api/members.html#add-a-member-to-a-group-or-project

### DIFF
--- a/NGitLab.Tests/MembersClientTests.cs
+++ b/NGitLab.Tests/MembersClientTests.cs
@@ -20,7 +20,7 @@ namespace NGitLab.Tests
             context.CreateNewUser(out var user);
             var projectId = project.Id.ToString(CultureInfo.InvariantCulture);
 
-            var expiresAt = DateTime.UtcNow.AddDays(30).ToString("yyyy-MM-dd");
+            var expiresAt = DateTimeOffset.UtcNow.AddDays(30).ToString("yyyy-MM-dd");
             context.Client.Members.AddMemberToProject(projectId, new ProjectMemberCreate
             {
                 AccessLevel = AccessLevel.Developer,
--- a/NGitLab.Tests/MembersClientTests.cs
+++ b/NGitLab.Tests/MembersClientTests.cs
@@ -1,14 +1,38 @@
-﻿using System.Globalization;
+﻿using System;
+using System.Globalization;
 using System.Linq;
 using System.Threading.Tasks;
 using NGitLab.Models;
 using NGitLab.Tests.Docker;
 using NUnit.Framework;
+#nullable enable
 
 namespace NGitLab.Tests
 {
     public class MembersClientTests
     {
+        [Test]
+        [NGitLabRetry]
+        public async Task AddMemberToProject()
+        {
+            using var context = await GitLabTestContext.CreateAsync();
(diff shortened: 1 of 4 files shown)


NGitLab.Tests/MembersClientTests.cs
```diff
@@ -1,14 +1,38 @@
...
+        [Test]
+        [NGitLabRetry]
+        public async Task AddMemberToProject()
+        {
+            using var context = await GitLabTestContext.CreateAsync();
+            var project = context.CreateProject();
+            context.CreateNewUser(out var user);
+            var projectId = project.Id.ToString(CultureInfo.InvariantCulture);
+
+            var expiresAt = DateTime.UtcNow.AddDays(30).ToString("yyyy-MM-dd");
```
louis-z: Use `DateTimeOffset` instead of `DateTime` (refer to this explanation of what the differences are: https://docs.microsoft.com/en-us/dotnet/standard/datetime/choosing-between-datetime).

NGitLab/Models/Membership.cs
```diff
@@ -41,5 +44,11 @@ public class Membership
         /// </summary>
         [DataMember(Name = "access_level")]
         public int AccessLevel;
+
+        /// <summary>
+        /// Membership expiration date
+        /// </summary>
+        [DataMember(Name = "expires_at")]
+        public DateTime? ExpiresAt;
```
louis-z, line 52: I have a few questions, knowing that this class inappropriately
- provides public fields rather than properties
- uses `DateTime`

At the risk of introducing breaking changes, should we
- convert fields to properties
- use `DateTimeOffset` throughout this class instead of `DateTime`?

@meziantou , what's your take on this? Can we afford that API change?

meziantou: We should avoid breaking changes if possible!

> provides public fields rather than properties

New classes use properties, older ones use fields. We try to not mix both in the same class but maybe there are some inconsistencies in the project.

SamuelTrottier: Is there a risk of breaking someone's code by switching the fields to properties if those properties all have getters and setters?

louis-z: Ok, so for now we'll keep using fields here... 😉  

However, the new field can probably be a `DateTimeOffset`. We use that type elsewhere and I'm fairly certain there is no JSON deserialization issues.

louis-z: > Is there a risk of breaking someone's code by switching the fields to properties if those properties all have getters and setters?

Unfortunately, yes. For now we'll keep using fields here... 😉  

However, the new field can probably be a `DateTimeOffset`. We use that type elsewhere and I'm fairly certain there is no JSON deserialization issues.

SamuelTrottier: Is it not better to keep it consistent with the `CreatedAt` Attribute?

louis-z: Yeah, ok. We'll revisit all `DateTime`s at a later time, when a planned refactoring is done.

PR #101, opened 2021-10-19
Implement Repository Tree in Mock

DomLatr:


NGitLab.Mock/Repository.cs
```diff
@@ -440,6 +440,21 @@ public Models.FileData GetFile(string filePath, string @ref)
...
+        {
+            var repo = GetGitRepository();
+            Commands.Checkout(repo, Project.DefaultBranch);
+
+            foreach(var fileSystemEntry in Directory.EnumerateFileSystemEntries(FullPath))
+            {
+                yield return new Models.Tree
+                {
+                    Name = Path.GetFileName(fileSystemEntry),
+                    Path = Path.GetFileName(fileSystemEntry),
```
Toa741, line 454: Type is likely:
- folder => tree
- file => blob

That should be easy to determine and set there :)

PR #103, opened 2021-10-20
Implement ProjectIssueNote mock

MarioSicaTeasdale:
Implement the ProjectIssueNote mocks.

### DIFF
--- a/NGitLab.Mock/Clients/ProjectIssueNoteClient.cs
+++ b/NGitLab.Mock/Clients/ProjectIssueNoteClient.cs
@@ -1,5 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
+
 using NGitLab.Models;
 
 namespace NGitLab.Mock.Clients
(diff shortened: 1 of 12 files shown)


NGitLab.Mock/Clients/ProjectIssueNoteClient.cs
```diff
@@ -14,24 +16,59 @@ public ProjectIssueNoteClient(ClientContext context, int projectId)
...
         }
 
-        public ProjectIssueNote Create(ProjectIssueNoteCreate create)
+        public Models.ProjectIssueNote Create(ProjectIssueNoteCreate create)
         {
-            throw new NotImplementedException();
+            using (Context.BeginOperationScope())
+            {
+                var project = GetProject(_projectId, ProjectPermission.Contribute);
+                var issue = project.Issues.First(iss => iss.Id == create.IssueId);
```
meziantou: I think you should use FirstOrDefault and throw `GitLabNotFoundException` if the value is null. This way it simulates returns a 404 error instead of throwing a _random_ exception.

Also, you can do similar changes in all other methods

meziantou (changes requested):

NGitLab.Mock/Clients/ProjectIssueNoteClient.cs
MarioSicaTeasdale: Thought it would be handled with the ArgumentNullException, but this is more readable, good idea. Done.

meziantou (approved):

PR #104, opened 2021-10-21
Add comments support in config mocks

LePtitDev:


NGitLab.Mock/Config/GitLabHelpers.cs
```diff
@@ -1301,9 +1496,46 @@ private static GitLabMergeRequest ToConfig(MergeRequest mergeRequest)
...
 
+            foreach (var comment in mergeRequest.Comments)
+            {
+                mrg.Comments.Add(ToConfig(comment));
+            }
+
             return mrg;
         }
 
+        private static GitLabComment ToConfig(ProjectIssueNote comment)
```
Toa741, line 1506: Is there a reason it's called `ToConfig` and not `ToComment` ?

Toa741:
@LePtitDev , do you have a link of the Api doc where you found the Comment properties?

NGitLab.Mock/Config/GitLabHelpers.cs
LePtitDev: Because it convert the comment model (here it's `ProjectIssueNote`) to a serializable/config representation for .yml file (see: https://github.com/ubisoft/NGitLab/blob/main/NGitLab.Mock/readme.md#3-with-yaml-file)

LePtitDev:
@Toa741, you can see API documentation here: https://docs.gitlab.com/ee/api/notes.html

PR #108, opened 2021-10-21
Add assignees support to mergeRequest mock

Bazooo:


NGitLab.Mock/Clients/MergeRequestClient.cs
```diff
@@ -585,6 +586,29 @@ public Models.MergeRequest Update(int mergeRequestIid, MergeRequestUpdate mergeR
                             throw new GitLabBadRequestException("user not found");
 
                         mergeRequest.Assignee = new UserRef(user);
+                        mergeRequest.Assignees.Add(new UserRef(user));
+                    }
+                }
+
+                if (mergeRequestUpdate.AssigneeIds != null)
```
louis-z, line 575: If both `mergeRequestUpdate.AssigneeId` and `mergeRequestUpdate.AssigneeIds` are set, it appears the latter will prevail. If that is indeed the GitLab behavior we want to emulate, it would be better to structure the code as follows so that no unnecessary processing is done:

```cs
if (mergeRequestUpdate.AssigneeIds != null)
{
    ...
}
else if (mergeRequestUpdate.AssigneeId != null)
{
    ...
}
```

louis-z (changes requested):

NGitLab.Mock/Clients/MergeRequestClient.cs
Bazooo: Good catch

louis-z (approved):

PR #117, opened 2021-10-28
Fix resolvable property not filled in model from mocks

LePtitDev:


LePtitDev:
I merged fixes in one PL #118

PR #124, opened 2021-11-04
Added support for confidential filter in issues

martin-bougie-ubi:
Added support for confidential filter when querying GitLab issues


NGitLab.Mock/Clients/IssueClient.cs
```diff
@@ -231,6 +231,11 @@ private IEnumerable<Issue> FilterByQuery(IEnumerable<Issue> issues, IssueQuery q
                 issues = issues.Where(i => string.Equals(i.Milestone?.Title, query.Milestone, StringComparison.Ordinal));
             }
 
+            if (query.Confidential != null)
+            {
+                issues = issues.Where(i => query.Confidential.Value || !i.Confidential);
```
meziantou: The filter should use the value of `query.Confidential`.

- query.Confidential == true => include all issues
- query.Confidential == false => include only non-confidential issues

In any case, if an issue is confidential, it should only be visible if the current user is `>= Reported` on the project or if the user is the author of the issue. (https://docs.gitlab.com/ee/user/project/issues/confidential_issues.html#permissions-and-access-to-confidential-issues)

meziantou (changes requested):

NGitLab.Mock/Clients/IssueClient.cs
martin-bougie-ubi: I added support for filtering on the user in the various gets of the mock. As for the filter, we made an incorrect assumption on the API behaviour. If the filter is set to true, then the API only returns confidential issues. If set to false, it only returns public issues. If the filter is not set, it returns both. We modified the mocks to reflect this behaviour.

NGitLab.Mock.Tests/NGitLab.Mock.Tests.csproj
```diff
@@ -1,7 +1,10 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="Current">
   <PropertyGroup>
     <TargetFrameworks>net5.0;net472</TargetFrameworks>
     <IsPackable>false</IsPackable>
+    <FileUpgradeFlags>40</FileUpgradeFlags>
```
meziantou: Revert

NGitLab.Mock/NGitLab.Mock.csproj
```diff
@@ -1,10 +1,13 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<Project>
+<Project ToolsVersion="Current">
   <Import Project="$([System.IO.Path]::Combine($(SolutionDir), '.SwitchToSource', $(SolutionFileName.Replace('.SwitchToSource.sln', '.sln')), 'import', 'NGitLab.Mock.before.props'))" Condition="$(SolutionFileName.EndsWith('.SwitchToSource.sln', StringComparison.OrdinalIgnoreCase)) And Exists($([System.IO.Path]::Combine($(SolutionDir), '.SwitchToSource', $(SolutionFileName.Replace('.SwitchToSource.sln', '.sln')), 'import', 'NGitLab.Mock.before.props')))" />
   <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />
   <PropertyGroup>
     <TargetFrameworks>net461;netstandard2.0</TargetFrameworks>
     <NoWarn>$(NoWarn);NU5104</NoWarn>
+    <FileUpgradeFlags>40</FileUpgradeFlags>
```
meziantou: revert

NGitLab/NGitLab.csproj
```diff
@@ -1,6 +1,9 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="Current">
   <PropertyGroup>
     <TargetFrameworks>net45;netstandard2.0</TargetFrameworks>
+    <FileUpgradeFlags>40</FileUpgradeFlags>
```
meziantou: revert

NGitLab.Mock/Clients/IssueClient.cs
```diff
@@ -106,7 +106,15 @@ public IEnumerable<Models.Issue> ForProject(int projectId)
         {
             using (Context.BeginOperationScope())
             {
-                return GetProject(projectId, ProjectPermission.View).Issues.Select(i => i.ToClientIssue()).ToList();
+                var project = GetProject(projectId, ProjectPermission.View);
+
+                return project
+                    .Issues
+                    .Where(i => project.CanUserViewConfidentialIssues(Context.User)
+                        || i.Author.Id == Context.User.Id
```
meziantou: I think you can create a new method `Issue.CanUserViewIssue(User)` and use it in all methods

NGitLab.Mock/Clients/IssueClient.cs
```diff
@@ -106,7 +106,15 @@ public IEnumerable<Models.Issue> ForProject(int projectId)
...
+                        || i.Assignee.Id == Context.User.Id)
```
meziantou: I don't think this is true, https://docs.gitlab.com/ee/user/project/issues/confidential_issues.html#permissions-and-access-to-confidential-issues

> However, a guest user can also create confidential issues, but can only view the ones that they created themselves.

meziantou (changes requested):

NGitLab.Mock/Issue.cs
```diff
@@ -95,5 +95,10 @@ public Models.Issue ToClientIssue()
                 Confidential = Confidential,
             };
         }
+
+        public bool CanUserViewIssue(User user)
+        {
+            return !Confidential || Author.Id == user.Id;
```
meziantou: Should include `Project.CanUserViewConfidentialIssues` so you can actually simplify all calls

NGitLab.Mock/Clients/IssueClient.cs
martin-bougie-ubi: I removed the free pass for assignees.

NGitLab.Mock/Issue.cs
martin-bougie-ubi: Good point, hadn't realized that the project object was already referenced inside the issue.

meziantou (approved):

PR #129, opened 2021-11-09
Added Advanced Search using Groups

weissm:
The  [Group API](https://docs.gitlab.com/ee/api/search.html#scope-projects-1) specifies the following format:

> `curl --header "PRIVATE-TOKEN: <your_access_token>" "https://gitlab.example.com/api/v4/groups/3/search?scope=projects&search=flight"`.

This group selection is supported by introducing the `groupID `parameter to the `ProjectQuery` scope.

### DIFF
--- a/NGitLab/Impl/ProjectClient.cs
+++ b/NGitLab/Impl/ProjectClient.cs
@@ -50,6 +50,11 @@ namespace NGitLab.Impl
                 url = $"/users/{query.UserId.Value.ToStringInvariant()}/projects";
             }
 
+            if (query.GroupID != string.Empty)
+            {
+                url = $"/groups/{query.GroupID}/search?scope=projects";
+            }
+
             if (query.LastActivityAfter.HasValue)
             {
                 url = Utils.AddParameter(url, "last_activity_after", query.LastActivityAfter.Value.UtcDateTime.ToString("O"));
--- a/NGitLab/Models/ProjectQuery.cs
+++ b/NGitLab/Models/ProjectQuery.cs
@@ -57,6 +57,11 @@ namespace NGitLab.Models
         public int? UserId;
 
         /// <summary>
+        /// Project visible by group
+        /// </summary>
+        public string GroupID;
+
+        /// <summary>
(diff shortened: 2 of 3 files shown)


louis-z:
Thanks for the contribution, @weissm. There are, however, a few things that need to be fixed:
- Since the route is `groups/3/search`, changes should be made to `GroupsClient` rather than `ProjectClient`.
- There are already a few search options offered by `GroupsClient`, e.g. [to search for a group](https://docs.gitlab.com/ee/api/groups.html#search-for-group) and [to list a group’s projects](https://docs.gitlab.com/ee/api/groups.html#list-a-groups-projects). We need to make it clear to the user what this new search feature offers.
- Maybe instead of using an existing Query type we should introduce a new `SearchParameters` type? At a glance, it seems that [all search types listed in the doc](https://docs.gitlab.com/ee/api/search.html) use the same set of parameters.
- Tests and mocks should be added as well

NGitLab/Models/ProjectQuery.cs
```diff
@@ -56,6 +56,11 @@ public class ProjectQuery
         /// </summary>
         public int? UserId;
 
+        /// <summary>
+        /// Project visible by group
+        /// </summary>
+        public string GroupID;
```
meziantou, line 62:
```suggestion
        public string GroupId;
```

NGitLab/Impl/ProjectClient.cs
```diff
@@ -50,6 +50,11 @@ public IEnumerable<Project> Get(ProjectQuery query)
                 url = $"/users/{query.UserId.Value.ToStringInvariant()}/projects";
             }
 
+            if (query.GroupID != string.Empty)
```
meziantou, line 53:
```suggestion
            if (!string.IsNullOrEmpty(query.GroupId))
```

meziantou (changes requested):
LGTM after changing the property name

weissm:
Incorporated comments in https://github.com/ubisoft/NGitLab/pull/131

PR #130, opened 2021-11-15
Added Advanced Search using Groups

weissm:
The  [Group API](https://docs.gitlab.com/ee/api/search.html#scope-projects-1) specifies the following format:

> `curl --header "PRIVATE-TOKEN: <your_access_token>" "https://gitlab.example.com/api/v4/groups/3/search?scope=projects&search=flight"`.

This group selection is supported by introducing the `groupID `parameter to the `ProjectQuery` scope.

It addresses the requests in the former pull request: `https://github.com/ubisoft/NGitLab/pull/129`. All tests are passing except for unrelated assert in `Test_merge_request_approvers`.

### DIFF
--- a/NGitLab/IGroupsClient.cs
+++ b/NGitLab/IGroupsClient.cs
@@ -29,6 +29,8 @@ namespace NGitLab
 
         IEnumerable<Project> SearchProjects(int groupId, string search);
 
+        IEnumerable<Project> SearchProjects(string groupId, string search, GroupQueryScope scope);
+
         void Delete(int id);
 
         Group Update(int id, GroupUpdate groupUpdate);
--- a/NGitLab/Impl/GroupsClient.cs
+++ b/NGitLab/Impl/GroupsClient.cs
@@ -91,6 +91,16 @@ namespace NGitLab.Impl
             return _api.Get().GetAll<Project>(url);
         }
 
+        // implementation according to https://docs.gitlab.com/ee/api/search.html#scope-projects-1
+        public IEnumerable<Project> SearchProjects(string groupId, string search, GroupQueryScope scope)
+        {
+            var url = Url + "/" + Uri.EscapeDataString(groupId.ToString(CultureInfo.InvariantCulture)) + "/search";
+            url = Utils.AddParameter(url, "scope", Enum.GetName(typeof(GroupQueryScope), scope).ToLower());
+            url = Utils.AddParameter(url, "search", search);
+            url = Utils.AddOrderBy(url);
+            return _api.Get().GetAll<Project>(url);
+        }
+
         public Group Create(GroupCreate group) => _api.Post().With(group).To<Group>(Url);
 
         public void Delete(int id)
--- a/NGitLab/Models/GroupQuery.cs
+++ b/NGitLab/Models/GroupQuery.cs
@@ -1,4 +1,5 @@
-﻿using System.Runtime.Serialization;
+﻿using System;
+using System.Runtime.Serialization;
 
 namespace NGitLab.Models
 {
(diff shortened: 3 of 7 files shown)


NGitLab/IGroupsClient.cs
```diff
@@ -29,6 +29,8 @@ public interface IGroupsClient
 
         IEnumerable<Project> SearchProjects(int groupId, string search);
 
+        IEnumerable<Project> SearchProjects(string groupId, string search, GroupQueryScope scope);
```
meziantou, line 32: When we add new overloads in the interface, we try to use a class. This way, we don't need to re-update it if we want to add a new parameter.
Also, as this is a new class, you should use properties (`{get;set}`) instead of fields.

```suggestion
        IEnumerable<Project> SearchProjects(SearchProjectQuery query);
```

NGitLab/Impl/GroupsClient.cs
```diff
@@ -91,6 +91,16 @@ public IEnumerable<Project> SearchProjects(int groupId, string search)
             return _api.Get().GetAll<Project>(url);
         }
 
+        // implementation according to https://docs.gitlab.com/ee/api/search.html#scope-projects-1
+        public IEnumerable<Project> SearchProjects(string groupId, string search, GroupQueryScope scope)
+        {
+            var url = Url + "/" + Uri.EscapeDataString(groupId.ToString(CultureInfo.InvariantCulture)) + "/search";
+            url = Utils.AddParameter(url, "scope", Enum.GetName(typeof(GroupQueryScope), scope).ToLower());
```
meziantou, line 98:
```suggestion
            url = Utils.AddParameter(url, "scope", scope.ToString().ToLowerInvariant());
```

NGitLab/IGroupsClient.cs
meziantou: According to the documentation examples, changing the `scope` also change the result type. For instance, `scope=merge_requests` returns a list of merge request. If so, we cannot allow any scope and return a list of projects.

NGitLab/Models/GroupQuery.cs
```diff
@@ -62,5 +63,56 @@ public class GroupQuery
         /// (optional)
         /// </summary>
         public AccessLevel? MinAccessLevel;
+
+        public GroupQueryScope Scope = GroupQueryScope.Projects;
```
meziantou, line 67: As it is required, let the user set it explicitly

```suggestion
        public GroupQueryScope? Scope;
```

NGitLab/NGitLab.csproj
```diff
@@ -2,6 +2,9 @@
   <PropertyGroup>
     <TargetFrameworks>net45;netstandard2.0</TargetFrameworks>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|net45|AnyCPU'">
```
meziantou, line 5: I think you should revert this change

meziantou (changes requested):

weissm:
Incorporated comments in https://github.com/ubisoft/NGitLab/pull/131

NGitLab/IGroupsClient.cs
weissm: Very valid comment. Shall we reduce now the GroupQueryScope options again to allow only the proper return types?

meziantou: If I'm not mistaken, the only valid option (to return a collection of projects) is `project`, but there is already this overload that does exactly that `IEnumerable<Project> SearchProjects(int groupId, string search);`

PR #137, opened 2021-11-22
Add async implementation

meziantou:
- Add a new dependency: `Microsoft.Bcl.AsyncInterfaces`
- Change the target framework from net45 to net461 as this is the minimum version supported by `Microsoft.Bcl.AsyncInterfaces`
- Add `GitLabCollectionResponse<T>`
- Add a unit test to ensure the new async methods follow best practices

fix #45

### DIFF
--- a/NGitLab.Tests/AsyncApiValidation.cs
+++ b/NGitLab.Tests/AsyncApiValidation.cs
@@ -19,6 +19,11 @@ namespace NGitLab.Tests
                 {
                     if (typeof(Task).IsAssignableFrom(method.ReturnType))
                     {
+                        if (!method.Name.EndsWith("Async", System.StringComparison.Ordinal))
+                        {
+                            Assert.Fail($"The method '{method}' must end with 'Async'");
+                        }
+
                         // Ensure method that returns a Task takes a CancellationToken
                         var parameterInfo = method.GetParameters().LastOrDefault();
                         if (parameterInfo is null)
@@ -41,6 +46,14 @@ namespace NGitLab.Tests
                         if (attr is null)
                         {
                             Assert.Fail($"The parameter '{parameterInfo.Name}' of '{method}' must be optional");
+                        }
+                    }
+
+                    if (method.ReturnType.IsGenericType && typeof(GitLabCollectionResponse<>).IsAssignableFrom(method.ReturnType.GetGenericTypeDefinition()))
+                    {
+                        if (!method.Name.EndsWith("Async", System.StringComparison.Ordinal))
+                        {
+                            Assert.Fail($"The method '{method}' must end with 'Async'");
                         }
                     }
                 }
--- a/NGitLab.Tests/AsyncApiValidation.cs
+++ b/NGitLab.Tests/AsyncApiValidation.cs
@@ -38,7 +38,7 @@ namespace NGitLab.Tests
 
                         if (!string.Equals(parameterInfo.Name, "cancellationToken", System.StringComparison.Ordinal))
                         {
-                            Assert.Fail($"The parameter '{parameterInfo.Name}' of '{method}' of must be named 'cancellationToken'");
+                            Assert.Fail($"The parameter '{parameterInfo.Name}' of '{method}' must be named 'cancellationToken'");
                         }
 
                         // Ensure the parameter is optional
--- a/NGitLab.Tests/AsyncApiValidation.cs
+++ b/NGitLab.Tests/AsyncApiValidation.cs
@@ -33,7 +33,7 @@ namespace NGitLab.Tests
 
                         if (parameterInfo.ParameterType != typeof(CancellationToken))
                         {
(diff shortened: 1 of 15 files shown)


NGitLab/IProjectClient.cs
```diff
@@ -25,6 +27,9 @@ public interface IProjectClient
         /// </summary>
         IEnumerable<Project> Get(ProjectQuery query);
 
+        // TODO We should not duplicate the method, but this would be a breaking change
+        GitLabCollectionResponse<Project> GetAll(ProjectQuery query);
```
meziantou: I think it would be better to make a breaking change and update the existing `Get` method in term of API. However, this can introduced lots of runtime errors as this is a binary breaking change.

I don't think there is a good way to handle this change...

NGitLab/Impl/IGitLabCollectionResponse.cs
```diff
@@ -0,0 +1,8 @@
+﻿using System.Collections.Generic;
+
+namespace NGitLab
+{
+    public interface IGitLabCollectionResponse<T> : IEnumerable<T>, IAsyncEnumerable<T>
```
jairbubbles: You want to keep is sync too?

meziantou: I wonder if this interface is useful. There are two locations where we need to instantiate the collections: NGitLab (actual HTTP requests) and NGitLab.Mocks (using a collection). We could remove the interface and support both scenarios using two ctors.

Note that exposing a type instead of `IAsyncEnumerable` is useful for the future. Indeed, we got requests to query all pages in parallel or to know the number of items (accessible in the headers of the first requests). Using a specific type would allow adding more properties / methods in the future.

jontdelorme:
I like the general approach. It makes it easy to add an async implementation to an interface

NGitLab.Tests/AsyncApiValidation.cs
```diff
@@ -0,0 +1,50 @@
...
+                        // Ensure method that returns a Task takes a CancellationToken
+                        var parameterInfo = method.GetParameters().LastOrDefault();
+                        if (parameterInfo is null)
+                        {
+                            Assert.Fail($"The method '{method}' must have a parameter of type 'CancellationToken'");
+                        }
+
+                        if (parameterInfo.ParameterType != typeof(CancellationToken))
+                        {
+                            Assert.Fail($"The parameter '{parameterInfo.Name}' of '{method}' of must be of type 'CancellationToken'");
```
louis-z: At this point `parameterInfo.Name` may not be equal to "cancellationToken". I would either validate the name first or adapt the message, something along the lines of
 ```suggestion
                            Assert.Fail($"The last parameter of method '{method}' must be of type 'CancellationToken' and named 'cancellationToken'");
```

File: NGitLab.Tests/AsyncApiValidation.cs [hunk 4]
```diff
@@ -0,0 +1,50 @@
...
+                        }
+
+                        if (!string.Equals(parameterInfo.Name, "cancellationToken", System.StringComparison.Ordinal))
+                        {
+                            Assert.Fail($"The parameter '{parameterInfo.Name}' of '{method}' of must be named 'cancellationToken'");
```
louis-z:
```suggestion
                            Assert.Fail($"The parameter '{parameterInfo.Name}' of '{method}' must be named 'cancellationToken'");
```

NGitLab/Impl/HttpRequestor.cs
```diff
@@ -54,6 +55,11 @@ public virtual void Execute(string tailAPIUrl)
             Stream(tailAPIUrl, parser: null);
         }
 
+        public virtual async Task ExecuteAsync(string tailAPIUrl, CancellationToken cancellationToken)
+        {
+            await StreamAsync(tailAPIUrl, parser: null, cancellationToken).ConfigureAwait(false);
```
Toa741: Is the use of async/await necessary here? Or we can return the task itself?

jairbubbles: Oh my, it's going to be a big debate. I believe there are two schools.

 I agree with you I feel like it makes the code simpler to return the task directly.

NGitLab/Impl/HttpRequestor.cs
```diff
@@ -54,6 +55,11 @@ public virtual void Execute(string tailAPIUrl)
...
+        }
```
meziantou:
```suggestion
        public virtual Task ExecuteAsync(string tailAPIUrl, CancellationToken cancellationToken)
        {
            return StreamAsync(tailAPIUrl, parser: null, cancellationToken);
        }
```

NGitLab/Impl/HttpRequestor.cs
meziantou: Fixed

Toa741 (approved):

PR #138, opened 2021-11-26
Add more async methods

meziantou:


NGitLab/Impl/PipelineClient.cs
```diff
@@ -48,14 +60,31 @@ public Job[] GetJobs(int pipelineId)
...
+        public GitLabCollectionResponse<Job> GetJobsAsync(PipelineJobQuery query)
         {
             var url = $"{_pipelinesPath}/{query.PipelineId.ToStringInvariant()}/jobs";
             url = Utils.AddParameter(url, "scope", query.Scope);
             url = Utils.AddParameter(url, "include_retried", query.IncludeRetried);
 
             // For some reason GitLab returns the jobs in the reverse order as
             // they appear in their UI. Here we reverse them!
-            return _api.Get().GetAll<Job>(url).Reverse();
+            return _api.Get().GetAllAsync<Job>(url);
```
louis-z, line 74: Missing call to `Reverse()`

NGitLab/Impl/RunnerClient.cs
```diff
@@ -94,5 +101,10 @@ public Runner Register(RunnerRegister request)
         {
             return _api.Post().With(request).To<Runner>(Runner.Url);
         }
+
+        public Task<Runner> EnableRunnerAsync(int projectId, RunnerId runnerId, CancellationToken cancellationToken = default)
+        {
+            throw new NotImplementedException();
```
louis-z: Missing implementation

NGitLab/IGroupsClient.cs
```diff
@@ -12,11 +14,15 @@ public interface IGroupsClient
 
         IEnumerable<Group> Search(string search);
 
+        GitLabCollectionResponse<Group> SearchAsync(string search);
```
louis-z, line 17: No CancellationToken?

Same for all other methods that return `GitLabCollectionResponse<T>`.

NGitLab/Impl/PipelineClient.cs
meziantou: I don't want to add the `Reverse` because:
- `Reverse` needs to get all results from the API before returning the first result to the caller
- NGitLab should just call the GitLab API, and should avoid changing the result

I think the non-async method should not have the `Reverse` call, but I don't know how people is using this method

NGitLab/Impl/RunnerClient.cs
meziantou: Added

NGitLab/IGroupsClient.cs
meziantou: Nope, there is no need for a `CancellationToken` here. You can set the `CancellationToken` when iterating on the collection (`await foreach(var item in collection.WithCancellation(token))`)

NGitLab/Impl/PipelineClient.cs
louis-z: Ok, then you need to remove the misleading comment that precedes the call.

You should also add a `<summary>` or `<remarks>` for the `GetJobsAsync` method, indicating that unlike the sync version it does not reverse the order.

NGitLab/IGroupsClient.cs
louis-z: Fair enough, but can you add documentation to that effect? When people see a method whose name ends with Async, they might expect it to take in a CancellationToken.

meziantou: Documentation added

PR #140, opened 2021-11-29
Add ProjectUpdate.TagList

meziantou:


NGitLab.Mock/Project.cs
```diff
@@ -355,6 +357,7 @@ public Models.Project ToClientProject()
                 ApprovalsBeforeMerge = ApprovalsBeforeMerge,
                 MergeMethod = MergeMethod,
                 Statistics = Statistics,
+                TagList = Tags,
```
louis-z, line 360: According to GitLab doc, `tag_list` has been deprecated in version 14.0 and should be replaced by `topics`:  
https://docs.gitlab.com/ee/api/projects.html

PR #141, opened 2021-11-30
Add mirror property in Project

Toa741:


louis-z:
Should we take this opportunity to add all mirror-related info?

image

PR #142, opened 2021-12-01
Bump up StyleCop.Analyzers to 1.2.0-beta.354 & fix some analyzer issues

louis-z:


NGitLab.Tests/Docker/GitLabTestContextRequestOptions.cs
```diff
@@ -180,7 +180,7 @@ private WebResponse LogRequest(HttpWebRequest request, WebResponse response)
         internal override Stream GetRequestStream(HttpWebRequest request)
         {
             var stream = new LoggableRequestStream(request.GetRequestStream());
-            _pendingRequest.AddOrUpdate(request, stream, (_, _) => stream);
+            _pendingRequest.AddOrUpdate(request, stream, (_, s) => s);
```
meziantou: I think you changed the behavior. `s` is the current value in the dictionary in case of an update. If you want to overwrite the value with `stream`, you should use the overload with `factoryArgument` ([doc](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.addorupdate?view=net-6.0#System_Collections_Concurrent_ConcurrentDictionary_2_AddOrUpdate__1__0_System_Func__0___0__1__System_Func__0__1___0__1____0_)) (or keep the existing code)

louis-z: Oops, you're right. Made the change too quickly. Was thrown off by MA0105. Maybe that rule needs to be adjusted for such a case?

louis-z: Reverted change.

meziantou (approved):

PR #144, opened 2021-12-10
Add squash option support in project model

LePtitDev:


NGitLab/Models/Project.cs
```diff
@@ -197,5 +197,8 @@ public class Project
 
         [DataMember(Name = "mirror_overwrites_diverged_branches")]
         public bool MirrorOverwritesDivergedBranches;
+
+        [DataMember(Name = "squash_option")]
+        public string SquashOption;
```
louis-z: The doc says that `squash_option` can be set to one of  the following: `never`, `always`, `default_on` or `default_off`. Perhaps it would make sense to define an enum for that. What do you think @LePtitDev?

LePtitDev: I preferred not to define an enumeration because if GitLab adds a new value, the deserialization will fail and I find that a shame for a secondary field.

LePtitDev: But I can make this enum if you prefer.

louis-z: I think you can get the best of both worlds by using `DynamicEnum`. It allows one to define a basic set of values (i.e. those we currently know about) but is robust enough to support any values that GitLab may add later on. Here's one such use:

https://github.com/ubisoft/NGitLab/blob/1f34721a9cbe09ddd389a02fc3bbedfa912cd4a1/NGitLab/Models/Event.cs#L23

PR #151, opened 2021-12-16
Add a few more async methods and tests

louis-z:
- Add `IProjectClient.GetByNamespacedPathAsync`
- Add `IRepositoryClient.GetTreeAsync`

### DIFF
--- a/NGitLab.Mock/Clients/ProjectClient.cs
+++ b/NGitLab.Mock/Clients/ProjectClient.cs
@@ -217,14 +217,10 @@ namespace NGitLab.Mock.Clients
             return GetById(id, query);
         }
 
-        public async Task<Models.Project> GetByFullNameAsync(string fullName, SingleProjectQuery query = null, CancellationToken cancellationToken = default)
+        public async Task<Models.Project> GetByNamespacedPathAsync(string path, SingleProjectQuery query = null, CancellationToken cancellationToken = default)
         {
             await Task.Yield();
-            using (Context.BeginOperationScope())
-            {
-                var project = GetProject(fullName, ProjectPermission.View);
-                return project.ToClientProject();
-            }
+            return this[path];
         }
 
         public IEnumerable<Models.Project> GetForks(string id, ForkedProjectQuery query)
--- a/NGitLab.Mock/Clients/ProjectClient.cs
+++ b/NGitLab.Mock/Clients/ProjectClient.cs
@@ -217,6 +217,16 @@ namespace NGitLab.Mock.Clients
             return GetById(id, query);
         }
 
+        public async Task<Models.Project> GetByFullNameAsync(string fullName, SingleProjectQuery query = null, CancellationToken cancellationToken = default)
+        {
+            await Task.Yield();
+            using (Context.BeginOperationScope())
+            {
+                var project = GetProject(fullName, ProjectPermission.View);
+                return project.ToClientProject();
+            }
+        }
+
         public IEnumerable<Models.Project> GetForks(string id, ForkedProjectQuery query)
         {
             using (Context.BeginOperationScope())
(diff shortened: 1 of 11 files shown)


NGitLab.Mock/Clients/ProjectClient.cs
```diff
@@ -217,6 +217,16 @@ public async Task<Models.Project> GetByIdAsync(int id, SingleProjectQuery query,
             return GetById(id, query);
         }
 
+        public async Task<Models.Project> GetByFullNameAsync(string fullName, SingleProjectQuery query = null, CancellationToken cancellationToken = default)
+        {
+            await Task.Yield();
+            using (Context.BeginOperationScope())
```
meziantou:
```suggestion
            return this[fullName];
```

NGitLab/IProjectClient.cs
```diff
@@ -56,6 +56,8 @@ public interface IProjectClient
 
         Task<Project> GetByIdAsync(int id, SingleProjectQuery query, CancellationToken cancellationToken = default);
 
+        Task<Project> GetByFullNameAsync(string fullName, SingleProjectQuery query = null, CancellationToken cancellationToken = default);
```
meziantou: I think we should use `GetByIdAsync` or `GetByPathWithNamespaceAsync`. FullName is confusing as it refers to the display name which could be different from the path.

meziantou: Maybe we should create a struct ProjectIdOrPathWithNamespace with implicit converters from string and int. This way have only one overload (not for this method as the other method already exists, but it could be useful for the following methods).

louis-z: Based on terminology found in the [GitLab doc](https://docs.gitlab.com/ee/api/index.html#namespaced-path-encoding), I went ahead with `GetByNamespacedPathAsync`.

louis-z: Having said that, I agree that
1. "Full Name" was incorrect... *though that's the name of the `ProjectClient` indexer's parameter* 
2. It would make sense to eventually introduce some kind of `ProjectIdOrPath` type

NGitLab.Mock/Clients/RepositoryClient.cs
```diff
@@ -60,6 +60,11 @@ public IEnumerable<Tree> GetTree(string path, string @ref, bool recursive)
             throw new NotImplementedException();
         }
 
+        public GitLabCollectionResponse<Tree> GetTreeAsync(RepositoryGetTreeOptions options)
+        {
+            throw new NotImplementedException();
```
meziantou: Should call `GetTree` as the sync implementation exist

NGitLab/Impl/RepositoryClient.cs
```diff
@@ -117,5 +109,25 @@ public IEnumerable<Commit> GetCommits(GetCommitsRequest request)
...
 
         public IProjectHooksClient ProjectHooks => new ProjectHooksClient(_api, _projectPath);
+
+        private string BuildGetTreeUrl(RepositoryGetTreeOptions options)
+        {
+            var url = $"{_repoPath}/tree";
+
+            var args = new List<string>(4);
+            if (!string.IsNullOrEmpty(options.Path))
+                args.Add($"path={options.Path}");
```
meziantou: Use `Utils.AddParameter` instead of creating a list of parameters

NGitLab/Impl/Utils.cs
```diff
@@ -52,6 +52,9 @@ public static string AddOrderBy(string url, string orderBy = null, bool supportK
 
         private static string AddParameterInternal(string url, string parameterName, string stringValue)
         {
+            if (string.IsNullOrEmpty(stringValue))
```
meziantou:
```suggestion
            if (stringValue is null)
```

For methods handling parameters such as `QueryHelpers.AddQueryString`, it's common to only consider null values
- null => don't add the parameter
- empty or any other value => add the parameter

louis-z: Oops, my bad, this is a remnant from a test. Reverted.

PR #154, opened 2021-12-21
Added get epic issues api route

apoufortin:
Added the missing route on the Epic API to get all issues associated to an epic.


NGitLab/IEpicClient.cs
```diff
@@ -21,6 +21,13 @@ public interface IEpicClient
         /// <returns></returns>
         Epic Get(int groupId, int epicId);
 
+        /// <summary>
+        /// Return all issues that are assigned to an epic
+        /// </summary>
+        /// <param name="groupId">Group ID</param>
+        /// <param name="epicId">Epic ID</param>
+        IEnumerable<Issue> GetIssues(int groupId, int epicId);
```
meziantou: Now that we have async implementation, you need to use it :)

```suggestion
        GitLabCollectionResponse<Issue> GetIssuesAsync(int groupId, int epicId);
```

NGitLab/Impl/EpicClient.cs
```diff
@@ -26,6 +27,11 @@ public Epic Get(int groupId, int epicId)
             return _api.Get().To<Epic>(string.Format(CultureInfo.InvariantCulture, SingleEpiceUrl, groupId, epicId));
         }
 
+        public IEnumerable<Issue> GetIssues(int groupId, int epicId)
+        {
+            return _api.Get().GetAll<Issue>(string.Format(CultureInfo.InvariantCulture, GroupEpicIssuesUrl, groupId, epicId));
```
meziantou:
```suggestion
            return _api.Get().GetAllAsync<Issue>(string.Format(CultureInfo.InvariantCulture, GroupEpicIssuesUrl, groupId, epicId));
```

meziantou (changes requested):

NGitLab/IEpicClient.cs
apoufortin: comme aucun autres des methodes sont async, ca ne semble pas trop uniforme de n'en mettre que une.  Ca va rend sont utilisation weird, non?

meziantou: Nope, it's the new way of doing it! Don't get fooled by the async suffix, `GitLabCollectionResponse` implements `IEnumerable` and `IAsyncEnumerable`, so the consumer can choose how they want to enumerate the result.

For the details, we had no choice to suffix the method name because we could not change existing methods. So, while the method ends with "Async", it is not actually async. It is async when you use `await foreach`.

meziantou (approved):

PR #157, opened 2022-01-04
Add CloneParameters on project mock

jairbubbles:
- Add new generic `CloneParameters` parameter that allows to specify `--filter:blob=none`
- Use `file://<local_path>` as remote url so that filter can be used to avoid 
```
--filter is ignored in local clones; use file:// instead.
```
- Set `uploadpack.allowFilter` to `true` in git config to avoid
```
warning: filtering not recognized by server, ignoring
```


NGitLab.Mock.Tests/NGitLab.Mock.Tests.csproj
```diff
@@ -1,4 +1,6 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
```
louis-z: Revert changes made to this file

louis-z (changes requested):

NGitLab.Mock.Tests/NGitLab.Mock.Tests.csproj
louis-z: In fact, while you're at it, do you think you could also make a similar reversion to NGitLab.Mock.csproj?

PR #160, opened 2022-01-10
Use System.Text.Json instead of SimpleJson

louis-z:
This entails the following:
- Add specific type converters for `enum`s, `DynamicEnum<T>`, `Sha1`, etc.
- Adapt API so that no JSON serialization types are exposed (as was the case for `GitLabException.ErrorObject`)
 
## Rationale for changing serializers
`System.Text.Json` is a modern serializer that offers much better performances than `SimpleJson`. For instance, using a simple benchmark that deserializes an array of 20 `Project`s, we get the following performances:

image

`System.Text.Json` is 5 times faster and allocates 100 times less memory.


NGitLab.Tests/BranchClientTests.cs
```diff
@@ -26,7 +26,7 @@ public async Task Test_CommitInfoIsCorrectlyDeserialized()
             Assert.LessOrEqual(7, commit.ShortId.Length);
 
             var fiveMinutesAgo = DateTime.UtcNow - TimeSpan.FromMinutes(5);
-            Assert.Less(fiveMinutesAgo, commit.CreatedAt);
+            Assert.Less(fiveMinutesAgo, commit.CreatedAt.ToUniversalTime());
```
meziantou: The current version doesn't return UTC date (`DateTime.Kind`)? Maybe we should fix that in another MR if needed

NGitLab/GitLabException.cs
```diff
@@ -37,8 +37,6 @@ public GitLabException(string message, Exception inner)
         /// </summary>
         public HttpStatusCode StatusCode { get; set; }
 
-        public JsonObject ErrorObject { get; set; }
```
meziantou, line 40: I think you can preserve this property (but changing the return type) by using `JsonExtensionDataAttribute`

NGitLab/Impl/Json/EnumConverter.cs
```diff
@@ -0,0 +1,79 @@
...
+using NGitLab.Models;
+
+namespace NGitLab.Impl.Json
+{
+    internal sealed class EnumConverter : JsonConverterFactory
+    {
+        public override bool CanConvert(Type type)
+        {
+            // AccessLevel will be serialized as int, using System.Text.Json's default enum serializer
+            return type.IsEnum && type != typeof(AccessLevel);
```
meziantou: Can't you use `JsonStringEnumConverter`?

NGitLab.Tests/BranchClientTests.cs
louis-z: IIRC the result of deserializing `"created_at"` differs
- `SimpleJson` converts it to UTC, while
- `System.Text.Json` keeps it in Local time

NGitLab/GitLabException.cs
louis-z: I always felt it was a mistake to leak out to the public API types that are associated to the serializer. Anything related to the serializer should remain internal.

Furthermore, has `ErrorObject` ever been useful to anyone? I think not, but if I'm wrong, I propose that we make it something more generic. Maybe `IDictionary<string, object>`?

NGitLab/Impl/Json/EnumConverter.cs
louis-z: AFAIK `JsonStringEnumConverter` does not handle the `[EnumMember]` attribute.

NGitLab.Tests/BranchClientTests.cs
meziantou: GitLab is inconsistent (once again). For an unknown reason it uses "+00:00" instead of "Z" for `created_at`...
So, S.T.Json converts it to a local date.

I would prefer a custom converter if needed to avoid potential issues with date comparison.

The test should be unchanged.

NGitLab/GitLabException.cs
meziantou: We do use this property to extract error details in some cases.
`IDictionary<string, object>` is fine

NGitLab/Impl/API.cs
```diff
@@ -64,7 +64,6 @@ private string OpenPrivateSession()
                 {
                     OriginalCall = new Uri(ex.OriginalCall.OriginalString.Replace(_credentials.Password, hiddenPassword)),
                     StatusCode = ex.StatusCode,
-                    ErrorObject = ex.ErrorObject,
```
meziantou: Keep the property and use `IDictionary<string, object?>`

NGitLab/Impl/Json/ReflectionExtensions.cs
```diff
@@ -0,0 +1,56 @@
...
+                    .Cast<EnumMemberAttribute>()
+                    .FirstOrDefault()?
+                    .Value;
+
+                yield return new EnumMapping((Enum)enumValue, stringValue);
+            }
+        }
+    }
+
+    internal struct EnumMapping
```
meziantou:
```suggestion
    internal readonly struct EnumMapping
```

NGitLab/Impl/Json/Serializer.cs
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        public static bool TryDeserializeObject(string json, out object obj)
+        {
+            try
+            {
+                obj = Deserialize<object>(json);
+                return true;
+            }
+            catch (Exception /*ex*/)
```
meziantou:
```suggestion
            catch
```

NGitLab/Models/MergeRequestApprove.cs
```diff
@@ -9,12 +10,14 @@ public class MergeRequestApprove
...
         /// </summary>
         [DataMember(Name = "sha")]
+        [JsonPropertyName("sha")]
         public string Sha { get; set; }
 
         /// <summary>
         /// (optional) - Current user’s password. Required if Require user password to approve is enabled in the project settings.
         /// </summary>
         [DataMember(Name = "approval_password ")]
+        [JsonPropertyName("approval_password ")]
```
meziantou:
```suggestion
        [JsonPropertyName("approval_password")]
```

NGitLab/PublicAPI.Unshipped.txt
```diff
@@ -2818,6 +2793,7 @@ NGitLab.Models.UserToken.CreatedAt.get -> System.DateTime
 NGitLab.Models.UserToken.CreatedAt.set -> void
 NGitLab.Models.UserToken.ExpiresAt.get -> System.DateTime?
 NGitLab.Models.UserToken.ExpiresAt.set -> void
+NGitLab.Models.UserToken.ExpiresAtStr.get -> string
```
meziantou: I think we can avoid these new props by using a custom converter for DateTime

NGitLab/PublicAPI.Unshipped.txt
```diff
@@ -2773,22 +2748,22 @@ NGitLab.Models.User.CanCreateGroup -> bool
 NGitLab.Models.User.CanCreateProject -> bool
 NGitLab.Models.User.ColorSchemeId -> int
 NGitLab.Models.User.CommitEmail -> string
-NGitLab.Models.User.ConfirmedAt -> System.DateTime
+NGitLab.Models.User.ConfirmedAt -> System.DateTime?
 NGitLab.Models.User.CreatedAt -> System.DateTime
-NGitLab.Models.User.CurrentSignIn -> System.DateTime
+NGitLab.Models.User.CurrentSignIn -> System.DateTime?
```
meziantou: Could we implement a custom converter to set the prop to `DateTime.MinValue` (current behavior) when the json prop is null?

NGitLab/PublicAPI.Unshipped.txt
```diff
@@ -2550,14 +2525,14 @@ NGitLab.Models.ResourceLabelEventAction.Add = 0 -> NGitLab.Models.ResourceLabelE
...
-NGitLab.Models.Runner.ContactedAt -> System.DateTime
+NGitLab.Models.Runner.ContactedAt -> System.DateTime?
 NGitLab.Models.Runner.Description -> string
 NGitLab.Models.Runner.Id -> int
 NGitLab.Models.Runner.IpAddress -> string
 NGitLab.Models.Runner.IsShared -> bool
 NGitLab.Models.Runner.Locked -> bool
 NGitLab.Models.Runner.Name -> string
-NGitLab.Models.Runner.Online -> bool
+NGitLab.Models.Runner.Online -> bool?
```
meziantou: Could we use a custom converter to avoid the breaking change?

NGitLab/PublicAPI.Unshipped.txt
```diff
@@ -1358,7 +1333,7 @@ NGitLab.Models.Event.TargetIId.get -> long?
 NGitLab.Models.Event.TargetIId.set -> void
 NGitLab.Models.Event.TargetTitle.get -> string
 NGitLab.Models.Event.TargetTitle.set -> void
-NGitLab.Models.Event.TargetType.get -> NGitLab.DynamicEnum<NGitLab.Models.EventTargetType>
+NGitLab.Models.Event.TargetType.get -> NGitLab.DynamicEnum<NGitLab.Models.EventTargetType>?
```
meziantou: I think a custom converter could map `null` to `EventTargetType.None` to avoid the breaking change

NGitLab.Tests/BranchClientTests.cs
louis-z: Added a custom converter & put back test as it was before.

NGitLab/GitLabException.cs
louis-z: Done

NGitLab/PublicAPI.Unshipped.txt
louis-z: Done

NGitLab/PublicAPI.Unshipped.txt
louis-z: Done

NGitLab/Impl/Json/BooleanConverter.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace NGitLab.Impl.Json
+{
+    internal sealed class BooleanConverter : JsonConverter<bool>
```
meziantou, line 7: Could you add some tests for these converters?

NGitLab/Impl/Json/Int64Converter.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace NGitLab.Impl.Json
+{
+    internal sealed class Int64Converter : JsonConverter<long>
```
meziantou, line 8: Could you add a test to validate this converter is not used for `long?`?

NGitLab/Impl/Json/Int64Converter.cs
```diff
@@ -0,0 +1,33 @@
...
+                return 0;
+
+            if (reader.TokenType == JsonTokenType.Number)
+                return reader.GetInt64();
+
+            // Change.AMode & BMode are deserialized from strings...
+            if (reader.TokenType == JsonTokenType.String)
+            {
+                var stringValue = reader.GetString();
+                if (long.TryParse(stringValue, out var value))
```
meziantou: CultureInvariant

NGitLab/Impl/Json/BooleanConverter.cs
louis-z: The converters were added to make existing tests pass. So, the converter code is already validated (and without it tests would fail). Do you still think we need to add more tests?

NGitLab/Impl/Json/Int64Converter.cs
louis-z: According to the [System.Text.Json documentation](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-converters-how-to?pivots=dotnet-6-0#handle-null-values), this custom converter will **not** be used for `Nullable<long>`:

```
By default, the serializer handles null values as follows:
- For reference types and Nullable<T> types:
  - It does not pass null to custom converters on serialization.
  - It does not pass JsonTokenType.Null to custom converters on deserialization.
  - It returns a null instance on deserialization.
  - It writes null directly with the writer on serialization.
- For non-nullable value types:
  - It passes JsonTokenType.Null to custom converters on deserialization. (If no custom converter is available, a JsonException exception is thrown by the internal converter for the type.)
```

So unless you think the documentation is lying 😁, I don't see the need for such a test.

What do you think?

NGitLab/Impl/Json/BooleanConverter.cs
meziantou: You don't need to validate them all, but I would prefer some tests to validate it correctly handle `null` values and also the parsing of datetime in different format (and that they are all UTC).

NGitLab/Impl/Json/Int64Converter.cs
meziantou: Let's trust the documentation!

NGitLab/Impl/Json/Int64Converter.cs
louis-z: Done

NGitLab/Impl/Json/BooleanConverter.cs
louis-z: I added some tests. Let me know if you feel there should be more.

NGitLab.Tests/Impl/JsonConverterTests.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        [TestCase("2022-01-12")]
+        [TestCase("2022-01-12T22:49:21.552Z")]
+        [TestCase("2022-01-12T22:49:21.552+00:00")]
+        public void Test_DeserializeStringToDateTime_SupportsMultipleFormats(string input)
+        {
+            var json = $@"{{ ""a_date_time"": ""{input}"" }}";
+            var obj = Serializer.Deserialize<MyDataContract>(json);
+            Assert.NotNull(obj);
+            Assert.AreEqual(DateTimeKind.Utc, obj.SomeDateTime.Kind);
```
meziantou: Should it be unknown for `[TestCase("2022-01-12")]`?

louis-z: Yeah, I guess that would make sense. I'll revisit the code...

louis-z: Done.

Also, I removed unnecessary `DateTime` format strings, i.e. those without which the tests still pass.

PR #161, opened 2022-01-11
Modify misleading parameter description

maximetinu:
Hi 👋

The description of this parameter made me thought it would be a commit ID, but it's not. It expects a branch name, I realised by reading these GitLab doc: https://docs.gitlab.com/ee/ci/triggers/#pass-cicd-variables-in-the-api-call

This is more of a suggestion but, would not it be better making this more explicit in the parameter description?

### DIFF
--- a/NGitLab/IPipelineClient.cs
+++ b/NGitLab/IPipelineClient.cs
@@ -54,7 +54,7 @@ namespace NGitLab
         /// <summary>
         /// Create a new pipeline.
         /// </summary>
-        /// <param name="ref">Reference to commit</param>
+        /// <param name="ref">Branch or tag</param>
         Pipeline Create(string @ref);
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


louis-z:
Actually, [the GitLab documentation](https://docs.gitlab.com/ee/api/pipelines.html#create-a-new-pipeline) does describe it as `Reference to commit`. However, you may be right in saying it must be a branch name or a tag, and not a sha1.

maximetinu:
So the description is inherited from them. But yeah I tried with a sha1, because I thought it was expecting that, but it didn't work.

I didn't tried with a tag, but it works well with branches. They also show it with a branch in their examples:

`curl --request POST --header "PRIVATE-TOKEN: <your_access_token>" "https://gitlab.example.com/api/v4/projects/1/pipeline?ref=main"`

maximetinu:
I've submitted a MR to the gitlab docs to propose this same modification, as it wouldn't make much sense for NGitLab to differ from them https://gitlab.com/gitlab-org/gitlab/-/merge_requests/77978 if they change it, I guess we'd be fine changing it as well

maximetinu:
They also believed that the change was correct, made a few grammar corrections and merged it, so I've updated this description to `The branch or tag to run the pipeline on` to reflect their change

PR #162, opened 2022-01-12
Add inheritedMembers to GetMemberOfProject

MarioSicaTeasdale:


NGitLab.Mock/Clients/MembersClient.cs
```diff
@@ -54,6 +54,12 @@ public Membership GetMemberOfProject(string projectId, string userId)
                 .FirstOrDefault(u => string.Equals(u.Id.ToString(CultureInfo.InvariantCulture), userId, StringComparison.Ordinal));
         }
 
+        public Membership GetMemberOfProject(string projectId, string userId, bool includeInheritedMembers)
+        {
+            return OfProject(projectId, includeInheritedMembers: true)
```
louis-z: Shouldn't you be passing the `includeInheritedMembers` parameter instead of hard-coding the value?

MarioSicaTeasdale: Woops lol good catch! Was testing it and forgot to remove it after ':)

PR #164, opened 2022-01-14
Fix WebResponse timing out

### LINKED ISSUE #163: Webresponse timing out when doing multiple requests since NGitLab 6.11.0
Since the version 6.11.0, there seems to be an issue where doing multiple requests (for functional tests for example) times out when getting the response. 

Versions before 6.11.0 worked perfectly fine.

mchanyeechoy:
I found that putting the yield outside the while loop fixes the timing out problem.

Fix #163

### DIFF
--- a/NGitLab/Impl/HttpRequestor.cs
+++ b/NGitLab/Impl/HttpRequestor.cs
@@ -152,6 +152,7 @@ namespace NGitLab.Impl
             public override async IAsyncEnumerator<T> GetAsyncEnumerator(CancellationToken cancellationToken = default)
             {
                 var nextUrlToLoad = _startUrl;
+                var buffer = new List<T>();
                 while (nextUrlToLoad != null)
                 {
                     var request = new GitLabRequest(nextUrlToLoad, MethodType.Get, data: null, _apiToken, _options.Sudo);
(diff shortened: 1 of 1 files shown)


NGitLab/Impl/HttpRequestor.cs
```diff
@@ -162,14 +163,17 @@ public override async IAsyncEnumerator<T> GetAsyncEnumerator(CancellationToken c
                     using var streamReader = new StreamReader(stream);
                     var responseText = await streamReader.ReadToEndAsync().ConfigureAwait(false);
                     var deserialized = SimpleJson.DeserializeObject<T[]>(responseText);
-                    foreach (var item in deserialized)
-                        yield return item;
+                    buffer.AddRange(deserialized);
```
meziantou: By using a list, you cannot stream the response. You'll enumerate all pages before returning the first result to the caller. 

I think you should try to first change the scope of the response / stream by using explicit using block (`using (...) {}`), so you ensure the response and stream are disposed before returning the items.

meziantou (changes requested):

meziantou:
Do you think it would be possible to add a test to ensure there is no regression on this one?

meziantou (approved):

PR #169, opened 2022-01-28
IiDs of mocked merge requests are equal to the IDs

### LINKED ISSUE #168: IiDs of mocked merge requests are equal to the IDs
In a GitLab server, IDs and IiDs are different, but mocked merge requests do not mimic that behavior. 

Say you're creeating a mocked MR:
- The id is set using `mergeRequest.Id = Server: Interlocked.Increment(ref _lastMergeRequestId);`
  -> ID = 1
- The IiD however uses a different logic: `mergeRequest.Id = MergeRequestsCollection: Select(mr => mr.Iid).DefaultIfEmpty().Max() + 1;`
  -> IiD = 1

As a solution, we could just keep incrementing the ID.

msrouchou:
close #168

### DIFF
--- a/NGitLab.Mock/GitLabServer.cs
+++ b/NGitLab.Mock/GitLabServer.cs
@@ -14,7 +14,7 @@ namespace NGitLab.Mock
     {
         private int _lastProjectId;
         private int _lastGroupId;
-        private int _lastMergeRequestId;
+        private int _lastMergeRequestId = 1000;
         private int _lastRunnerId;
         private int _lastIssueId;
         private int _lastMilestoneId;
--- a/NGitLab.Mock/GitLabServer.cs
+++ b/NGitLab.Mock/GitLabServer.cs
@@ -12,16 +12,17 @@ namespace NGitLab.Mock
     /// </summary>
     public sealed class GitLabServer : GitLabObject, IDisposable
     {
-        private int _lastProjectId;
-        private int _lastGroupId;
-        private int _lastMergeRequestId = 1000;
-        private int _lastRunnerId;
-        private int _lastIssueId;
-        private int _lastMilestoneId;
-        private int _lastPipelineId;
-        private int _lastJobId;
-        private int _lastBadgeId;
-        private int _lastLabelId;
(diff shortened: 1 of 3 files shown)


NGitLab.Mock/MergeRequestCollection.cs
```diff
@@ -45,7 +45,7 @@ public override void Add(MergeRequest mergeRequest)
 
             if (mergeRequest.Iid == default)
             {
-                mergeRequest.Iid = GetNewIid();
+                mergeRequest.Iid = Server.GetNewMergeRequestId();
```
meziantou: In a project, `Iid` are incremented values => This logic doesn't consider the project, so it cannot be valid.

As mentioned in https://github.com/ubisoft/NGitLab/issues/168#issuecomment-1024642870, I think we should just change the initial value of `_lastMergeRequestId`. This way `IId` will be incremented values but will always be different from the `Id`.

Changing the initial `Id` should not impact anyone as there is no reason to use the `Id` property in the context of the GitLab API (this is a deprecated info)

meziantou (changes requested):

NGitLab.Mock/MergeRequestCollection.cs
meziantou: To be clear, I think we should set a default value for all these fields (could be random or a high value such as 10_000) https://github.com/ubisoft/NGitLab/blob/main/NGitLab.Mock/GitLabServer.cs#L15-L24

NGitLab.Mock/GitLabServer.cs
```diff
@@ -14,7 +14,7 @@ public sealed class GitLabServer : GitLabObject, IDisposable
     {
         private int _lastProjectId;
         private int _lastGroupId;
-        private int _lastMergeRequestId;
+        private int _lastMergeRequestId = 1000;
```
meziantou: Do the same for all fields. Also, add a comment to explain why we do it.

meziantou (approved):

PR #174, opened 2022-02-09
Added issue Time tracking statistics route

apoufortin:
Can you create a version once this is merged?

Thanks.


NGitLab/Impl/IssueClient.cs
```diff
@@ -87,5 +90,10 @@ public IEnumerable<MergeRequest> ClosedBy(int projectId, int issueIid)
         {
             return _api.Get().GetAll<MergeRequest>(string.Format(CultureInfo.InvariantCulture, ClosedByUrl, projectId, issueIid));
         }
+
+        public Task<TimeStats> TimeStats(int projectId, int issueIid, CancellationToken cancellationToken = default)
```
meziantou: The method name must end with "Async", that's why the test "ValidateAsyncMethodSignature" failed

```suggestion
        public Task<TimeStats> TimeStatsAsync(int projectId, int issueIid, CancellationToken cancellationToken = default)
```

PR #175, opened 2022-02-10
Add graphql support

meziantou:


NGitLab.Tests/GraphQLTests.cs
```diff
@@ -0,0 +1,67 @@
...
+                Variables =
+                {
+                    ["path"] = project.PathWithNamespace,
+                },
+            });
+
+            Assert.AreEqual("gid://gitlab/Project/" + project.Id, response.Project.Id);
+        }
+
+        private sealed class ProjectReponse
```
louis-z, line 55: You're missing an 's': `ProjectResponse`

NGitLab.Mock/Clients/GitLabClient.cs
```diff
@@ -37,6 +39,12 @@ public GitLabClient(ClientContext context)
 
         public IMergeRequestClient MergeRequests => new MergeRequestClient(Context);
 
+        public IGraphQLClient GraphQL
+        {
+            get => _overrideGraphQLClient ?? new GraphQLClient(Context);
```
louis-z, line 44: Couldn't we store the new instance if `_overrideGraphQLClient` was null?
`get => _overrideGraphQLClient ??= new GraphQLClient(Context);`

meziantou: We could do it for all similar properties. You can do it in another PR if you want :)

NGitLab.Tests/GraphQLTests.cs
meziantou: I'm not sure, there is only one project in the response

louis-z: I'm not talking about singular vs plural forms, I meant you spelled the word `Reponse` wrong.
It should have been spelled Re**s**ponse.

meziantou: I should use a spell checker...

meziantou: Fixed in https://github.com/ubisoft/NGitLab/pull/178

PR #178, opened 2022-02-11
Add DefaultGraphQLClient to configure a global GraphQL client

meziantou:


NGitLab.Mock/Clients/GitLabClient.cs
```diff
@@ -41,7 +41,7 @@ public GitLabClient(ClientContext context)
 
         public IGraphQLClient GraphQL
         {
-            get => _overrideGraphQLClient ?? new GraphQLClient(Context);
+            get => _overrideGraphQLClient ?? Server.DefaultGraphQLClient ?? new GraphQLClient(Context);
```
louis-z: I'm not sure I like the idea of having 3 different locations for the GraphQL client... Can you at least store it in the field?
```suggestion
            get => _overrideGraphQLClient ??= Server.DefaultGraphQLClient ?? new GraphQLClient(Context);
```

meziantou: In the server makes it easier to setup for most usages. Per client allows to simulate different permissions. The default case is not relevant. It's just to avoid `NullReferenceException`s.

`Can you at least store it in the field` => as you closed your PR, I'll do it here.

meziantou: done

PR #179, opened 2022-02-11
Fix misspelled word & store GraphQL client in field

louis-z:
Fixes for
- https://github.com/ubisoft/NGitLab/pull/175#discussion_r804609829
- https://github.com/ubisoft/NGitLab/pull/175#discussion_r804611589

### DIFF
--- a/NGitLab.Mock/Clients/GitLabClient.cs
+++ b/NGitLab.Mock/Clients/GitLabClient.cs
@@ -2,7 +2,7 @@
 {
     internal sealed class GitLabClient : ClientBase, IGitLabClient
     {
-        private IGraphQLClient _overrideGraphQLClient;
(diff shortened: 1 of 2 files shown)


NGitLab.Mock/Clients/GitLabClient.cs
```diff
@@ -41,8 +41,8 @@ public GitLabClient(ClientContext context)
 
         public IGraphQLClient GraphQL
         {
-            get => _overrideGraphQLClient ?? new GraphQLClient(Context);
-            internal set => _overrideGraphQLClient = value;
+            get => _graphQLClient ??= new GraphQLClient(Context);
```
meziantou, line 44: I think we should do it for all properties or not do it. I don't think there is a reason to special-case this client.

PR #183, opened 2022-02-14
Fix converters for pipeline.coverage

meziantou:


NGitLab/Impl/Json/DoubleConverter.cs
```diff
@@ -11,6 +12,12 @@ public override double Read(ref Utf8JsonReader reader, Type type, JsonSerializer
             if (reader.TokenType == JsonTokenType.Null)
                 return 0.0;
 
+            if (reader.TokenType == JsonTokenType.String)
+            {
+                if (double.TryParse(reader.GetString(), NumberStyles.Any, CultureInfo.InvariantCulture, out var d))
+                    return (int)d;
+            }
```
louis-z, line 19: - Why cast to `int`?
- Is `NumberStyles.Any` better than `NumberStyles.None` here?
- At any rate, can you make the code consistent with the existing one?

https://github.com/ubisoft/NGitLab/blob/07da8863a70d663a6777597b633b9dc2793450ab/NGitLab/Impl/Json/Int64Converter.cs#L19-L24

meziantou: > Why cast to int?

A bad copy paste...

> Is NumberStyles.Any better than NumberStyles.None here?

We must use Any to support negative numbers. Fixed in https://github.com/ubisoft/NGitLab/pull/184

PR #184, opened 2022-02-14
Fix test for converters

meziantou:


louis-z:
Can you address all bullet points from https://github.com/ubisoft/NGitLab/pull/183#discussion_r806011550?

louis-z:
I don't think we need `NullableInt32Converter`, as by default the serializer handles null values as follows for nullable value types:

- It does not pass `null` to custom converters on serialization.
- It does not pass `JsonTokenType.Null` to custom converters on deserialization.
- It returns a `null` instance on deserialization.
- It writes `null` directly with the writer on serialization.

Cf. https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-converters-how-to?pivots=dotnet-6-0#handle-null-values

Am I missing something?

PR #186, opened 2022-02-16
Add async methods on IssueClient

Toa741:
- On ResourceLabelEvents, I suppose it was an issueIid instead of issueId, as on other methods and as the description of the param suggested it.

### DIFF
--- a/NGitLab/Impl/IssueClient.cs
+++ b/NGitLab/Impl/IssueClient.cs
@@ -30,9 +30,19 @@ namespace NGitLab.Impl
             return _api.Get().GetAll<Issue>(string.Format(CultureInfo.InvariantCulture, ProjectIssuesUrl, projectId));
         }
 
+        public GitLabCollectionResponse<Issue> ForProjectAsync(int projectId)
+        {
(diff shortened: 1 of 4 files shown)


NGitLab/Impl/IssueClient.cs
```diff
@@ -30,70 +30,126 @@ public IEnumerable<Issue> ForProject(int projectId)
...
+        {
+            return _api.Get().GetAllAsync<MergeRequest>(string.Format(CultureInfo.InvariantCulture, ClosedByUrl, projectId, issueIid));
+        }
+
         public Task<TimeStats> TimeStatsAsync(int projectId, int issueIid, CancellationToken cancellationToken = default)
         {
             return _api.Get().ToAsync<TimeStats>(string.Format(CultureInfo.InvariantCulture, TimeStatsUrl, projectId, issueIid), cancellationToken);
         }
+
+        private IEnumerable<Issue> Get(string url, IssueQuery query)
```
meziantou: I'm not sure you need this method. `GitLabCollectionResponse` implements `IEnumerable<T>`, so you may only need the other overload

PR #188, opened 2022-02-18
ReviewerIds on Merge Request Create

Yekonori:


NGitLab/Models/MergeRequestCreate.cs
```diff
@@ -22,6 +22,10 @@ public class MergeRequestCreate
         [JsonPropertyName("assignee_ids")]
         public int[] AssigneeIds;
 
+        [DataMember(Name = "reviewer_ids")]
+        [JsonPropertyName("reviewer_ids")]
+        public int[] ReviewerIds;
```
meziantou, line 27: You need to add the new member to [PublicAPI.Unshipped.txt](https://github.com/ubisoft/NGitLab/blob/main/NGitLab/PublicAPI.Unshipped.txt) to fix the CI.
You can use the quick fix in Visual Studio

image

meziantou (commented):

You need to fix the warning to fix the CI. Then, I can merge the PR.

PR #192, opened 2022-02-21
add noteable metadata to note

Codex04:
https://docs.gitlab.com/ee/api/events.html

### DIFF
--- a/NGitLab/Models/Note.cs
+++ b/NGitLab/Models/Note.cs
@@ -46,5 +46,13 @@ namespace NGitLab.Models
         [DataMember(Name = "confidential")]
         [JsonPropertyName("confidential")]
         public bool Confidential;
+
+        [DataMember(Name = "noteable_iid")]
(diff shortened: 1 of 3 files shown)


NGitLab/Models/Note.cs
```diff
@@ -46,5 +46,13 @@ public class Note
         [DataMember(Name = "confidential")]
         [JsonPropertyName("confidential")]
         public bool Confidential;
+
+        [DataMember(Name = "noteable_iid")]
+        [JsonPropertyName("noteable_iid")]
+        public long NoteableIid { get; set; }
```
louis-z, line 52: Since [`noteable` is actually a misspelling of `notable`](https://en.wiktionary.org/wiki/noteable), maybe we could rename our properties appropriately?

louis-z: On second thought, seeing as there are already occurrences of that misspelling in NGitLab, we should probably stick with it for the sake of consistency. 😢

PR #199, opened 2022-03-03
Remove UriFix, needed in net45 but now oxbow code

louis-z:


meziantou:
How did you validate the change as the tests currently only validate net5? (manual?)

Maybe we should multi-target net48 to be sure NGitLab works on .NET Framework.

louis-z:
Before removing the UriFix test, I
- modified the project to target `net472` (which meant commenting out a bunch of `net5.0`-specific stuff...)
- modified the test to instantiate a `Uri` directly

The test ran fine.

PR #200, opened 2022-03-03
Add support to retrieve last activity date for all users

louis-z:
Add support for [`/user/activities`](https://docs.gitlab.com/ee/api/users.html#get-user-activities-admin-only)

### DIFF
--- a/NGitLab/IUserClient.cs
+++ b/NGitLab/IUserClient.cs
@@ -61,6 +61,7 @@ namespace NGitLab
 
         /// <summary>
         /// Gets the last activity date for all users, sorted from oldest to newest
+        /// (<seealso href="https://docs.gitlab.com/ee/api/users.html#get-user-activities-admin-only">GitLab documentation</seealso>).
         /// </summary>
         /// <param name="from">Date from which activities will be considered. If unspecified, will look back over the last 6 months.</param>
         /// <remarks>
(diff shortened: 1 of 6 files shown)


NGitLab/IUserClient.cs
```diff
@@ -57,5 +58,21 @@ public interface IUserClient
...
+        /// The activities that update the timestamp are:
+        /// <list type="bullet">
+        /// <item><description>Git HTTP/SSH activities (such as clone, push)</description></item>
+        /// <item><description>User logging in to GitLab</description></item>
+        /// <item><description>User visiting pages related to dashboards, projects, issues, and merge requests (introduced in GitLab 11.8)</description></item>
+        /// <item><description>User using the API</description></item>
+        /// <item><description>User using the GraphQL API</description></item>
+        /// </list>
+        /// </remarks>
+        IEnumerable<LastActivityDate> GetLastActivityDates(DateTimeOffset? from = null);
```
meziantou: New methods should provide async calls. `GitLabCollectionResponse` implements `IEnumerable` and `IAsyncEnumerable`, so it is the way to go for collections. Also, you need to add the `Async` suffix for consistency (even if it could be synchronous)

```suggestion
        GitLabCollectionResponse<LastActivityDate> GetLastActivityDatesAsync(DateTimeOffset? from = null);
```

NGitLab/IUserClient.cs
```diff
@@ -57,5 +58,21 @@ public interface IUserClient
         /// Get a list of users that match the query.
         /// </summary>
         IEnumerable<User> Get(UserQuery query);
+
+        /// <summary>
+        /// Gets the last activity date for all users, sorted from oldest to newest
```
meziantou, line 63:
```suggestion
        /// Gets the last activity date for all users, sorted from oldest to newest.
        /// <seealso href="https://docs.gitlab.com/ee/api/users.html#get-user-activities-admin-only">GitLab documentation</seealso>
```

meziantou (changes requested):

meziantou (approved):

PR #212, opened 2022-03-28
Add support for bridge jobs

kagouball:
Add support for bridge jobs
Pipeline bridges doc : https://docs.gitlab.com/ee/api/jobs.html#list-pipeline-bridges

### DIFF
--- a/NGitLab/IPipelineClient.cs
+++ b/NGitLab/IPipelineClient.cs
@@ -107,6 +107,6 @@ namespace NGitLab
         /// Returns the bridges of a pipeline.
         /// </summary>
         /// <param name="query"></param>
-        GitLabCollectionResponse<Bridge> GetBridgesAsync(PipelineJobQuery query);
+        GitLabCollectionResponse<Bridge> GetBridgesAsync(PipelineBridgeQuery query);
     }
 }
--- a/NGitLab/IPipelineClient.cs
+++ b/NGitLab/IPipelineClient.cs
@@ -102,5 +102,18 @@ namespace NGitLab
         /// </summary>
         /// <param name="pipelineId">ID of the pipeline</param>
         TestReport GetTestReports(int pipelineId);
+
+        /// <summary>
+        /// Returns the bridges of a pipeline.
+        /// </summary>
+        /// <param name="query"></param>
+        IEnumerable<Bridge> GetBridges(PipelineJobQuery query);
+
+
+        /// <summary>
+        /// Returns the bridges of a pipeline.
+        /// </summary>
+        /// <param name="query"></param>
+        GitLabCollectionResponse<Bridge> GetBridgesAsync(PipelineJobQuery query);
     }
 }
--- a/NGitLab/IPipelineClient.cs
+++ b/NGitLab/IPipelineClient.cs
@@ -107,13 +107,6 @@ namespace NGitLab
         /// Returns the bridges of a pipeline.
         /// </summary>
         /// <param name="query"></param>
-        IEnumerable<Bridge> GetBridges(PipelineJobQuery query);
-
-
-        /// <summary>
-        /// Returns the bridges of a pipeline.
-        /// </summary>
-        /// <param name="query"></param>
-        GitLabCollectionResponse<Bridge> GetBridgesAsync(PipelineJobQuery query);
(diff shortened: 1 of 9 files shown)


kagouball:
Mocks are coming

NGitLab/Models/Job.cs
```diff
@@ -1,47 +1,14 @@
...
-        [DataMember(Name = "stage")]
-        [JsonPropertyName("stage")]
-        public string Stage;
-
-        [DataMember(Name = "coverage")]
-        [JsonPropertyName("coverage")]
-        public double? Coverage;
+        [DataMember(Name = "pipeline")]
+        [JsonPropertyName("pipeline")]
+        public JobPipeline Pipeline;
```
Toa741: `pipeline` property seems also available on bridges

NGitLab/Models/Bridge.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Runtime.Serialization;
+using System.Text.Json.Serialization;
+
+namespace NGitLab.Models
+{
+    [DataContract]
+    public class Bridge : JobCommon
+
+    {
+        [DataMember(Name = "downstream_pipeline")] [JsonPropertyName("downstream_pipeline")]
```
Toa741:
```suggestion
        [DataMember(Name = "downstream_pipeline")]
        [JsonPropertyName("downstream_pipeline")]
```

NGitLab/IPipelineClient.cs
```diff
@@ -102,5 +102,18 @@ public interface IPipelineClient
         /// </summary>
         /// <param name="pipelineId">ID of the pipeline</param>
         TestReport GetTestReports(int pipelineId);
+
+        /// <summary>
+        /// Returns the bridges of a pipeline.
+        /// </summary>
+        /// <param name="query"></param>
+        IEnumerable<Bridge> GetBridges(PipelineJobQuery query);
```
meziantou: This method can be removed. `GetBridgesAsync` can be enumerated synchronously. Indeed, `GitLabCollectionResponse` implements both `IEnumerable` and `IAsyncEnumerable`.

NGitLab/Models/JobCommon.cs
```diff
@@ -0,0 +1,94 @@
+﻿using System;
+using System.Runtime.Serialization;
+using System.Text.Json.Serialization;
+
+namespace NGitLab.Models
+{
+    [DataContract]
+    public class JobCommon
```
meziantou: We should match the GitLab naming (https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/api/entities/ci/job_basic.rb)

```suggestion
    public class JobBasic
```

meziantou: Adding a base entity and moving fields and nested class is a binary breaking change. 

@jairbubbles Do you think this is a problem? If not, should we replace the fields by properties?

jairbubbles: I don't think it's too problematic, the jobs API is probably not used in many places. Yes, let's switch to properties in the meantime.

NGitLab/IPipelineClient.cs
jairbubbles: Indeed, I forgot about it and the "Async" was misleading. We should rename it to "AsyncOrNot" 😛

kagouball:
Mocks  in next PR

NGitLab/Impl/PipelineClient.cs
```diff
@@ -189,5 +189,25 @@ public TestReport GetTestReports(int pipelineId)
...
+        {
+            var url = CreateGetBridgesUrl(query);
+            return _api.Get().GetAllAsync<Bridge>(url);
+        }
+
+        private string CreateGetBridgesUrl(PipelineJobQuery query)
+        {
+            var url = $"{_pipelinesPath}/{query.PipelineId}/bridges";
+            url = Utils.AddParameter(url, "scope", query.Scope);
+            url = Utils.AddParameter(url, "include_retried", query.IncludeRetried);
```
Toa741: From the doc, `include_retried` seems not available for bridges query :/

jairbubbles: Let's make a `PipelineBridgeQuery` then!

PR #216, opened 2022-04-07
Add Downstream Pipeline Mocks

kagouball:


NGitLab.Mock/Pipeline.cs
```diff
@@ -12,7 +12,9 @@ public Pipeline(string @ref)
             Ref = @ref;
         }
 
-        public new Project Parent => (Project)base.Parent;
+        public Project Project => Parent is Pipeline pipeline ? pipeline.Project : (Project)Parent;
+
+        public bool IsParentAPipeline { get; set; }
```
jairbubbles: I'm not too convinced by `IsParentAPipeline`, it looks like a hack could you keep it internal at least?

NGitLab.Mock/Config/GitLabHelpers.cs
```diff
@@ -1422,7 +1424,16 @@ private static void CreatePipeline(GitLabServer server, Project project, GitLabP
                 Status = pipeline.Jobs.Count == 0 ? JobStatus.NoBuild : JobStatus.Success,
             };
 
-            project.Pipelines.Add(ppl);
+            if (parent is Project project)
```
meziantou, line 1427: You should use something different than `Parent` to track the parent pipeline. I'm pretty sure it will break elsewhere... Adding a new property or type would be better. For instances:
- `Pipeline.Bridges`
- `Server.PipelineBridges`

jairbubbles: Keep in mind that bridges are fake jobs in a pipeline that points to their own (downstream) pipeline. Here we are creating a downstream pipeline.

meziantou: So, you are creating a new pipeline (already supported) + a new information to track the downstream/upstream pipeline. The parent is not the right property to change. It should be a new property in the upstream/downstream pipeline (called it bridge / downstreamPipeline / upstreamPipeline / whatever), or a new collection in the project/server.

But the parent property is used elsewhere and I'm pretty sure it will break existing code to set it to a pipeline instead of a Project

jairbubbles: Ok then we're on the same page!

Downstream pipelines look like they are quite normal pipelines but they are not returned when you list project pipelines (`GET /projects/:id/pipelines`) and using `source=parent_pipeline` does not seem to work. But they are returned when you get a single pipeline with the downstream pipeline's Id (`GET /projects/:id/pipelines/:pipeline_id`). So that's why we add them to the projects anyway like normal pipelines.

I guess we can just add an internal `IsDownStreamPipeline` property and filter in the different routes and don't change the parent. I don't think it's important to know what is the parent pipeline. We currently use the bridges to get to the downstream pipelines.

NGitLab.Mock/Pipeline.cs
jairbubbles: After discussing it more in the other thread, I would just rename it to IsDownStreamPipeline.

PR #220, opened 2022-04-11
Add implementation of get single issue

Ditchfuuu:
This is an implementation of [Get Single Issue](https://docs.gitlab.com/ee/api/issues.html#single-issue) allowing admin access tokens to retrieve any issue by its unique ID. Happy to update anything as required, I was also unsure on the best approach to add tests for the new methods so have omitted them pending guidance if you'd like them adding.

### DIFF
--- a/NGitLab.Mock.Tests/IssuesMockTests.cs
+++ b/NGitLab.Mock.Tests/IssuesMockTests.cs
@@ -55,5 +55,21 @@ namespace NGitLab.Mock.Tests
             var client = server.CreateClient();
             Assert.DoesNotThrow(() => client.Issues.Get(new IssueQuery { Scope = "assigned_to_me" }).ToArray());
         }
+
+        [Test]
+        public void Test_issue_by_id_can_be_found()
+        {
+            using var server = new GitLabConfig()
+                .WithUser("user", isDefault: true, isAdmin: true)
+                .WithProject("Test", configure: project => project
+                    .WithIssue("Issue title", author: "user", id: 5))
+                .BuildServer();
+
+            var client = server.CreateClient();
+
+            var issue = client.Issues.GetById(10001);
+            Assert.AreEqual(5, issue.IssueId);
+            Assert.AreEqual("Issue title", issue.Title);
+        }
     }
 }
--- a/NGitLab/IIssueClient.cs
+++ b/NGitLab/IIssueClient.cs
@@ -56,6 +56,16 @@ namespace NGitLab
         GitLabCollectionResponse<Issue> GetAsync(int projectId, IssueQuery query);
 
         /// <summary>
+        ///     <para>Return a single issue, only for administrators.</para>
+        ///     <para>url like GET /issues/:id</para>
+        /// </summary>
+        /// <param name="issueId"></param>
+        /// <returns>The issue that corresponds to the issue id</returns>
+        Issue GetById(int issueId);
+
+        Task<Issue> GetByIdAsync(int issueId, CancellationToken cancellationToken = default);
+
+        /// <summary>
         /// Add an issue witht he proposed title to the GitLab list for the selected proejct id.
         /// </summary>
         /// <param name="issueCreate"></param>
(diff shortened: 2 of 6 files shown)


NGitLab/IIssueClient.cs
```diff
@@ -55,6 +55,16 @@ public interface IIssueClient
 
         GitLabCollectionResponse<Issue> GetAsync(int projectId, IssueQuery query);
 
+        /// <summary>
+        ///     <para>Return a single issue, only for administrators.</para>
+        ///     <para>url like GET /issues/:id</para>
+        /// </summary>
+        /// <param name="issueId"></param>
+        /// <returns>The issue that corresponds to the issue id</returns>
+        Issue GetById(int issueId);
```
Toa741, line 64: Is there a reason why not calling the method `Get` like the existing one?

Toa741: Bonus: can you update the method `Issue Get(int projectId, int issueId);` => It's an issueIid (`Issue Get(int projectId, int issueIid);`

Toa741:
For the tests, it should be fairly straightforward as the new method is quite close to the existing one to get an issue for a project:

- Add a test in `IssueTests`, there are examples. The context also have an AdminClient if required
- The same for the mocks, it should be really close to `Get(IssueQuery query)`, you'll normally just need to change the query condition :)

NGitLab/IIssueClient.cs
Ditchfuuu: Thanks for the feedback, regarding your first comment I received the following warning when I originally tried adding an overload for `GetAsync`:

> Warning RS0026 Symbol 'GetAsync' violates the backcompat requirement: 'Do not add multiple overloads with optional parameters'. See 'https://github.com/dotnet/roslyn/blob/main/docs/Adding%20Optional%20Parameters%20in%20Public%20API.md' for details.

I was unsure how to tackle this and was reluctant to introduce an issue so went with an entirely different method name, if there's something I can do to resolve this let me know and I'll update it.

I'll get to adding those tests 😃 !

Toa741: Oh ok, it's probably due to `GetAsync(int projectId, int issueId, CancellationToken cancellationToken = default);`

I suppose that if you don't put the default value to the cancellation token it will pass. `Task<Issue> GetByIdAsync(int issueId, CancellationToken cancellationToken);`

But on the rest of the methods we always pass `= default`

Do you have a suggestion @meziantou for that kind of case?

meziantou: I prefer `GetById` :) 

This is more explicit. Also, we use this convention in other clients.

Toa741: Ok :)

Toa741: @Ditchfuuu , you can concentrate on the Bonus then :)

Ditchfuuu: I've updated the `Get` as requested (hopefully!) and left `GetById` untouched.

Ditchfuuu:
I've added the tests following your guidance, they pass locally. I've tried to keep them in with the same style as the others but please let me know if I need to tweak them.

NGitLab.Mock.Tests/IssuesMockTests.cs
```diff
@@ -55,5 +55,21 @@ public void Test_issues_assignee_not_throwing_when_assignees_is_null()
...
             Assert.DoesNotThrow(() => client.Issues.Get(new IssueQuery { Scope = "assigned_to_me" }).ToArray());
         }
+
+        [Test]
+        public void Test_issue_by_id_can_be_found()
+        {
+            using var server = new GitLabConfig()
+                .WithUser("user", isDefault: true, isAdmin: true)
+                .WithProject("Test", configure: project => project
+                    .WithIssue("Issue title", author: "user", id: 5))
```
Toa741, line 65: Can you maybe use the WithIssue with the configure action overload?

To see if you can retrieve the internal Id (`10001` looks like it's out of nowhere :))

Ditchfuuu: Aha, I forgot to ask about this :) - I was unsure how to go about either setting or retrieving the internal ID. If I use the configure overload I think I can only set the IID and don't seem to have any access to the internal ID 😞. I can see it's in `client.Issues.Owned` so could retrieve the internal ID from there?

Toa741: I thought it would have been easier to get it, indeed it seems quite hidden :).

PR #221, opened 2022-04-11
Add blobs search from gitlab search Api

Toa741:
- https://docs.gitlab.com/ee/api/search.html#advanced-search-api
- Update GitLabDockerVersion in tests as the comment suggested
- Unable to add tests (scoped search requires elasticsearch to be installed separately)
- Not sure mocks can be easily implemented either

### DIFF
--- a/NGitLab/ISearchClient.cs
+++ b/NGitLab/ISearchClient.cs
@@ -4,6 +4,6 @@ namespace NGitLab
 {
     public interface ISearchClient
     {
-        GitLabCollectionResponse<SearchBlob> GetBlobs(SearchQuery query);
+        GitLabCollectionResponse<SearchBlob> GetBlobsAsync(SearchQuery query);
     }
 }
--- a/NGitLab/ISearchClient.cs
+++ b/NGitLab/ISearchClient.cs
@@ -1,0 +1,9 @@
+﻿using NGitLab.Models;
+
+namespace NGitLab
+{
+    public interface ISearchClient
+    {
+        GitLabCollectionResponse<SearchBlob> GetBlobs(SearchQuery query);
+    }
+}
--- a/NGitLab/Impl/SearchClient.cs
+++ b/NGitLab/Impl/SearchClient.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using NGitLab.Models;
+﻿using NGitLab.Models;
 
 namespace NGitLab.Impl
 {
--- a/NGitLab/Impl/SearchClient.cs
+++ b/NGitLab/Impl/SearchClient.cs
@@ -3,7 +3,7 @@ using NGitLab.Models;
 
 namespace NGitLab.Impl
 {
-    public class SearchClient : ISearchClient
+    internal sealed class SearchClient : ISearchClient
     {
         private readonly API _api;
         private readonly string _baseUrl;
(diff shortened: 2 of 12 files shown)


NGitLab/Impl/SearchClient.cs
```diff
@@ -0,0 +1,77 @@
+﻿using System;
+using NGitLab.Models;
+
+namespace NGitLab.Impl
+{
+    public class SearchClient : ISearchClient
```
meziantou:
```suggestion
    internal sealed class SearchClient : ISearchClient
```

NGitLab/ISearchClient.cs
```diff
@@ -0,0 +1,9 @@
+﻿using NGitLab.Models;
+
+namespace NGitLab
+{
+    public interface ISearchClient
+    {
+        GitLabCollectionResponse<SearchBlob> GetBlobs(SearchQuery query);
```
meziantou: I think other `GitLabCollectionResponse` methods use the `Async` suffix.
```suggestion
        GitLabCollectionResponse<SearchBlob> GetBlobsAsync(SearchQuery query);
```

NGitLab/Impl/SearchClient.cs
```diff
@@ -0,0 +1,77 @@
...
+            url = Utils.AddParameter(url, "search", query.Search);
+
+            if (query.PerPage.HasValue)
+            {
+                url = Utils.AddParameter(url, "per_page", query.PerPage);
+            }
+
+            if (query.Confidential.HasValue)
+            {
+                if (!string.Equals(scope, "issues", StringComparison.Ordinal))
```
meziantou: Can't we let GitLab do the check?

NGitLab/Impl/SearchClient.cs
```diff
@@ -0,0 +1,77 @@
...
+                {
+                    throw new NotSupportedException("Only allowed value for OrderBy is 'created_at'");
+                }
+
+                url = Utils.AddOrderBy(url, query.OrderBy);
+            }
+
+            if (!string.IsNullOrEmpty(query.Sort))
+            {
+                if (!string.Equals(query.Sort, "asc", StringComparison.Ordinal) &&
```
meziantou: I don't think we need a validation here. Let's GitLab do the job.

NGitLab/SearchQuery.cs
```diff
@@ -0,0 +1,37 @@
+﻿namespace NGitLab
+{
+    public class SearchQuery
```
meziantou:
```suggestion
    public sealed class SearchQuery
```

NGitLab/Impl/SearchClient.cs
Toa741: Yes, it will just ignore if not of the right type

NGitLab/PublicAPI.Unshipped.txt
```diff
@@ -669,8 +675,8 @@ NGitLab.Impl.RunnerClient.Register(NGitLab.Models.RunnerRegister request) -> NGi
 NGitLab.Impl.RunnerClient.RunnerClient(NGitLab.Impl.API api) -> void
 NGitLab.Impl.RunnerClient.this[int id].get -> NGitLab.Models.Runner
 NGitLab.Impl.RunnerClient.Update(int runnerId, NGitLab.Models.RunnerUpdate runnerUpdate) -> NGitLab.Models.Runner
+NGitLab.Impl.SearchClient.GetBlobs(NGitLab.SearchQuery query) -> NGitLab.GitLabCollectionResponse<NGitLab.Models.SearchBlob>
```
meziantou:
```suggestion
NGitLab.Impl.SearchClient.GetBlobsAsync(NGitLab.SearchQuery query) -> NGitLab.GitLabCollectionResponse<NGitLab.Models.SearchBlob>
```

PR #225, opened 2022-04-22
Adding IssueType supported in the API

apoufortin:


NGitLab/Models/IssueType.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace NGitLab.Models
+{
+    public enum IssueType
+    {
+        issue,
```
louis-z, line 5: Please use PascalCase names.

PascalCase <-> snake_case serialization should work correctly when there is only 1 word. This means you'll need to provide  `[EnumMember(Value = "test_case")]` attribute for `TestCase`.

louis-z (changes requested):

NGitLab/Models/IssueType.cs
louis-z: NVM, what I said is only partially right and only for JSON-serialized `enum`s. Here we're dealing with query params...

PR #226, opened 2022-04-26
Migrate tests from net5.0 to net6.0

louis-z:
Since .NET 5's End of Support is scheduled for May 8, 2022 (see [here](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core#lifecycle)), let's migrate to .NET 6 right away.


Toa741:
Should we actually try to migrate to using HttpClient? @meziantou 

Not sure it would be easily feasible though

louis-z:
> Should we actually try to migrate to using HttpClient? @meziantou
> 
> Not sure it would be easily feasible though

Well, I found a way: [I deleted the `NGitLab.Tests/HttpTracing.cs` file](https://github.com/ubisoft/NGitLab/pull/226/commits/ca37bc292ec6b42f15fa00662b3b36b57e8d6c33).

meziantou:
> Should we actually try to migrate to using HttpClient? @meziantou

I've already tried. As most methods are synchronous, it would not be great (sync over async). Starting with .NET HttpClient provides synchronous methods, but we target older frameworks. Also, it requires breaking changes on `RequestOptions`.

PR #230, opened 2022-05-13
Add issue due date to create and edit models

Ditchfuuu:
This allows `due_date` to be specified at issue creation time or when updating, as well as adding a test for this functionality. Let me know if anything needs tweaking.

### DIFF
--- a/NGitLab.Tests/IssueTests.cs
+++ b/NGitLab.Tests/IssueTests.cs
@@ -235,15 +235,13 @@ namespace NGitLab.Tests
             using var context = await GitLabTestContext.CreateAsync();
             var project = context.CreateProject();
             var issuesClient = context.Client.Issues;
-            const string formatString = "yyyy-MM-dd";
 
-            var initialDueDate = DateTime.Now.AddDays(3).ToString(formatString);
+            var initialDueDate = new DateTime(2022, 1, 1);
             var issue1 = issuesClient.Create(new IssueCreate { Id = project.Id, Title = "title1", DueDate = initialDueDate });
             var issue = issuesClient.Get(project.Id, issue1.IssueId);
-            Assert.IsNotNull(issue.DueDate);
-            Assert.AreEqual(initialDueDate, issue.DueDate?.ToString(formatString));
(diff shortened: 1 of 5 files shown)


NGitLab/Models/IssueCreate.cs
```diff
@@ -27,5 +27,8 @@ public class IssueCreate
 
         [JsonPropertyName("confidential")]
         public bool Confidential;
+
+        [JsonPropertyName("due_date")]
+        public string DueDate;
```
meziantou: Should be `DateTime`. You may need a custom json converter (`DateOnlyJsonConverter`) like https://github.com/ubisoft/NGitLab/blob/866ec15e8a98eecab7d6b5c892943b42a5509add/NGitLab/Impl/Json/DateTimeConverter.cs#L21 to serialize using `yyyy-MM-dd` format. Before implementing the converter, you should validate if the GitLab API accepts the ISO format.

NGitLab.Tests/IssueTests.cs
```diff
@@ -226,5 +227,32 @@ public async Task Test_get_all_resource_label_events()
...
+        [Test]
+        [NGitLabRetry]
+        public async Task Test_get_new_and_updated_issue_with_duedate()
+        {
+            using var context = await GitLabTestContext.CreateAsync();
+            var project = context.CreateProject();
+            var issuesClient = context.Client.Issues;
+            const string formatString = "yyyy-MM-dd";
+
+            var initialDueDate = DateTime.Now.AddDays(3).ToString(formatString);
```
meziantou: I don't think you need to use variable date (DateTime.Now). Instead, you can hardcode a date using `new DateTime(2022, 01, 01)` to reduce the chance of flakiness.

Ditchfuuu: For some reason I just assumed the API would reject a date in the past 🤦 updated to use hardcoded dates.

NGitLab/Models/IssueCreate.cs
Ditchfuuu: The ISO format didn't seem to work, so I've added a new `DateOnlyConverter` which seems to behave as expected.

PR #235, opened 2022-05-16
Correctly set the Tree path in NGitLab.Mocks

DomLatr:


NGitLab.Mock/Repository.cs
```diff
@@ -463,19 +463,48 @@ internal IEnumerable<Models.Tree> GetTree(Models.RepositoryGetTreeOptions reposi
...
                 Name = Path.GetFileName(filePath),
-                Path = Path.GetFileName(filePath),
-                Type = fileAttribute == FileAttributes.Directory ? Models.ObjectType.tree : Models.ObjectType.blob,
+                Path = GetTreeRelativePath(filePath),
+                Type = fileAttribute.HasFlag(FileAttributes.Directory) ? Models.ObjectType.tree : Models.ObjectType.blob,
             };
+
+            string GetTreeRelativePath(string fileFullPath)
+            {
+                if (!repositoryPath.EndsWith("\\", StringComparison.Ordinal))
```
louis-z: This will not work on Linux or macOS. You should probably use `Path.DirectorySeparatorChar` instead.

louis-z (changes requested):

NGitLab.Mock/Repository.cs
louis-z: In fact, can you add tests to validate the changes?

Issue #243, opened 2022-06-09
Add suggested and actual approvers to MergeRequestApprovals


seguemark:
The GitLab v4 api provides information about suggested approvers (based on approval rules) in the MergeRequestApproval API, using the "approved_by" and "suggested_approvers" properties. It also has an "approved" boolean. Those should be made available in the NGitLab.Models.MergeRequestApprovals object.

PR #244, opened 2022-06-09
Adds "approved", "approved_by" and "suggested_approvers" to the Merge…

seguemark:
…RequestApprovals model  (#243)

### DIFF
--- a/NGitLab/Models/MergeRequestApprovals.cs
+++ b/NGitLab/Models/MergeRequestApprovals.cs
@@ -7,6 +7,12 @@ namespace NGitLab.Models
         [JsonPropertyName("approvers")]
         public MergeRequestApprover[] Approvers;
 
+        [JsonPropertyName("approved_by")]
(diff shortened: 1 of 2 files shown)


NGitLab/Models/MergeRequestApprovals.cs
```diff
@@ -7,6 +7,12 @@ public class MergeRequestApprovals
         [JsonPropertyName("approvers")]
         public MergeRequestApprover[] Approvers;
 
+        [JsonPropertyName("approved_by")]
+        public MergeRequestApprover[] ApprovedBy { get; set; }
```
louis-z, line 11: You need to add this to the public API definition. You can use Visual Studio's "Quick Actions and Refactorings..." to quickly do it for all additions:

image

seguemark: Updated, thanks.

PR #245, opened 2022-06-16
Support multi files in Snippet

dranc:
Snippet now handle multi files
This is not properly explained in the doc
https://docs.gitlab.com/ee/api/snippets.html
But the api that get the file return an array of files (like in the post/update call).

Here is an example of what I receive :

```
{
    "id": 2694,
    "title": "TITLE",
    "description": "DESCRIPTION",
    "visibility": "internal",
    "updated_at": "2022-06-16T12:50:47.646Z",
    "created_at": "2022-06-15T12:05:51.654Z",
    "project_id": 1,
    "web_url": "https://gitlab-xxxxx/-/snippets/2694",
    "raw_url": "https://gitlab-xxxxx/-/snippets/2694/raw",
    "ssh_url_to_repo": "git@gitlab-xxxxx/snippets/2694.git",
    "http_url_to_repo": "https://gitlab-xxxxx/snippets/2694.git",
    "author": {
        "id": 1,
        "name": "xxxx",
        "username": "dranc",
        "state": "active",
        "avatar_url": "https://gitlab-xxxxx/uploads/-/system/user/avatar/1/avatar.png",
        "web_url": "https://gitlab-xxxxx/dranc"
    },
    "file_name": "ResourcesHelper.cs",
    "files": [
        {
            "path": "ResourcesHelper.cs",
            "raw_url": "https:/gitlab-xxxxxx/-/snippets/1/raw/main/file1.cs"
        },
        {
            "path": "SecondFile.txt",
            "raw_url": "https:/gitlab-xxxxx/-/snippets/1/raw/main/file2.txt"
        }
    ]
}
```

I've added it.


meziantou:
Thanks for your contribution!
 Could you add an automated test to validate your changes?

dranc:
Hey,
Don't really know how to test a model but I've added a test anyway.
Tell me if it need more. :)

NGitLab.Tests/SnippetsTest.cs
```diff
@@ -73,5 +74,20 @@ public async Task Test_snippet_inProject(VisibilityLevel visibility)
 
             snippetClient.Delete(newSnippet.ProjectId, returnedProjectSnippet.Id);
         }
+
+        [Test]
+        public void Test_snippet_deserialize()
```
meziantou: I would expect a test that uses `snippetClient.Create`/`snippetClient.Get` or another method of `snippetClient` that returns a list of files. You can assert the list is not empty and contains the expected files.

NGitLab.Tests/SnippetsTest.cs
```diff
@@ -75,8 +75,45 @@ public async Task Test_snippet_inProject(VisibilityLevel visibility)
...
+                Files = new[]
+                {
+                    new SnippetCreateFile() { Action = SnippetCreateFile.ActionType.Create, FilePath = "Path1.txt", Content = "Content1" },
+                    new SnippetCreateFile {  Action = SnippetCreateFile.ActionType.Create, FilePath = "Path2.txt", Content = "Content2" },
+                },
+            };
+
+            // act - assert
+            snippetClient.Create(newSnippet1);
+            var snippet = snippetClient.All.First();
```
meziantou: Tests can run in parallel, so you cannot be sure the snippet you get is the one you created in this test. It could be the one created by another test.

You can use the same logic as in the first test: https://github.com/ubisoft/NGitLab/pull/245/files/75f9baa1d6534cd2efab70bc70503267d0085bb0..bd08bfab21b604fb61bd9a3a86a74a6eb945dede#diff-2fe4fab06a5114d750bd61926c57591d0ed237feec7282522c740a10d28fa9bcR38

NGitLab.Tests/SnippetsTest.cs
```diff
@@ -75,8 +75,45 @@ public async Task Test_snippet_inProject(VisibilityLevel visibility)
...
+
+            Assert.IsNotNull(snippet.Files);
+            Assert.IsNotEmpty(snippet.Files);
+            Assert.IsTrue(snippet.Files.Length == 2);
```
meziantou: To get better error message, you can use `Assert.AreEqual`

```suggestion
            Assert.AreEqual(2, snippet.Files.Length);
```

NGitLab.Tests/SnippetsTest.cs
```diff
@@ -75,8 +75,45 @@ public async Task Test_snippet_inProject(VisibilityLevel visibility)
...
+                    new SnippetCreateFile {  Action = SnippetCreateFile.ActionType.Create, FilePath = "Path2.txt", Content = "Content2" },
+                },
+            };
+
+            // act - assert
+            snippetClient.Create(newSnippet1);
+            var snippet = snippetClient.All.First();
+
+            Assert.IsNotNull(snippet.Files);
+            Assert.IsNotEmpty(snippet.Files);
```
meziantou: This assert is already included in the next assert. So, you can remove it as it doesn't provide any additional value.

```suggestion
```

NGitLab.Tests/SnippetsTest.cs
```diff
@@ -75,8 +75,45 @@ public async Task Test_snippet_inProject(VisibilityLevel visibility)
...
         }
 
+        [TestCase(VisibilityLevel.Private)]
+        [TestCase(VisibilityLevel.Internal)]
+        [TestCase(VisibilityLevel.Public)]
+        [NGitLabRetry]
+        public async Task Test_snippet_files(VisibilityLevel visibility)
+        {
+            using var context = await GitLabTestContext.CreateAsync();
+            var project = context.CreateProject();
```
meziantou: The test doesn't use this project instance. I think you can remove this line.

NGitLab.Tests/SnippetsTest.cs
```diff
@@ -73,5 +74,57 @@ public async Task Test_snippet_inProject(VisibilityLevel visibility)
...
+            Assert.IsNotNull(snippet.Files);
+            Assert.IsNotEmpty(snippet.Files);
+            Assert.IsTrue(snippet.Files.Length == 2);
+
+            Assert.IsNotNull(snippet.Files[0]);
+            Assert.IsTrue(string.Equals(snippet.Files[0].Path, "Path1.txt", StringComparison.Ordinal));
+        }
+
+        [Test]
+        public void  Test_snippet_deserialize()
```
meziantou: I'm not sure this test is very useful as the other one already tests deserialization.

NGitLab/Models/SnippetCreateFile.cs
```diff
@@ -0,0 +1,48 @@
+﻿using System.ComponentModel.DataAnnotations;
+using System.Runtime.Serialization;
+using System.Text.Json.Serialization;
+
+namespace NGitLab.Models
+{
+    public class SnippetCreateFile
+    {
+        public enum ActionType
```
meziantou: This enum and the `PreviousFile` property are only allowed to [update a snippet](https://docs.gitlab.com/ee/api/snippets.html#update-snippet). So, it should not be part of the `SnippetCreateFile` class.

The `Create` method uses the `POST` verb, whereas the `Update` (non-existant in NGitLab now) uses `PUT`. So, we cannot use the same method to create and update a snippet. 
I think, we should add a new method to update a snippet to the client, and create a new class `SnippetUpdate` class with this enum and other properties. This could be part of another PR if you prefer.

meziantou (changes requested):

NGitLab/Models/SnippetCreateFile.cs
dranc: Ok, yes this will be a bigger change ! ;)
I will try to do it today or tomorrow.

NGitLab/Models/SnippetUpdateFile.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System.ComponentModel.DataAnnotations;
+using System.Runtime.Serialization;
+using System.Text.Json.Serialization;
+
+namespace NGitLab.Models
+{
+    public class SnippetUpdateFile : SnippetCreateFile
```
meziantou: Create and Update are different concepts (even if they share some properties now). They can evolve differently in the future and new properties in Create may not be valid in Update.

So, we should not use a base class.

NGitLab/Models/SnippetUpdateFile.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        public enum ActionType
```
meziantou: Public nested type is not encouraged ([CA1034](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1034)). This project has some legacy, so you'll find wrong usages.

For new types we should follow best practices. You can move the enum out of the parent type and rename it to something more meaningful like `SnippetUpdateFileAction`

NGitLab/Models/SnippetUpdate.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.ComponentModel.DataAnnotations;
+using System.Text.Json.Serialization;
+
+namespace NGitLab.Models
+{
+    public class SnippetUpdate
+    {
+        [Required]
+        [JsonPropertyName("id")]
+        public int SnippedId;
```
meziantou: Use getter and setter for new properties.

```suggestion
        public int SnippetId { get; set; }
```

NGitLab/Models/SnippetProjectUpdate.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace NGitLab.Models
+{
+    public class SnippetProjectUpdate
+    {
+        public int ProjectId;
+
+        [Required]
+        [JsonPropertyName("id")]
+        public int SnippedId;
```
meziantou:
```suggestion
        public int SnippetId { get; set; }
```

NGitLab/Models/SnippetProjectCreate.cs
```diff
@@ -5,10 +5,11 @@ namespace NGitLab.Models
 {
     public class SnippetProjectCreate
     {
-        [Required]
-        [JsonPropertyName("id")]
         public int ProjectId;
 
+        [JsonPropertyName("id")]
+        public int SnippedId;
```
meziantou: The Create method doesn't have a SnippetId field:  https://docs.gitlab.com/ee/api/project_snippets.html#create-new-snippet

dranc: In fact it does when the value is returned after we create it but we don't use it so I remove it anyway...
image

NGitLab/Models/SnippetProjectUpdate.cs
```diff
@@ -10,7 +10,7 @@ public class SnippetProjectUpdate
 
         [Required]
         [JsonPropertyName("id")]
-        public int SnippedId;
+        public int SnippedId { get; set; }
```
meziantou:
```suggestion
        public int SnippetId { get; set; }
```

NGitLab/Models/SnippetUpdate.cs
```diff
@@ -8,7 +8,7 @@ public class SnippetUpdate
     {
         [Required]
         [JsonPropertyName("id")]
-        public int SnippedId;
+        public int SnippedId { get; set; }
```
meziantou:
```suggestion
        public int SnippetId { get; set; }
```

.idea/.idea.NGitLab/.idea/.gitignore
```diff
@@ -0,0 +1,13 @@
+# Default ignored files
```
meziantou: Could you remove the `.idea` files from this PR?

NGitLab/Models/SnippetProjectUpdate.cs
meziantou: The diff still show Snippe**d**Id instead SnippetId. Maybe you forgot to push some changes.

NGitLab/Models/SnippetUpdate.cs
dranc: Wow sorry, did not see the typo ! :/

meziantou (approved):
LGTM when the PublicAPI file is fixed!

https://github.com/ubisoft/NGitLab/runs/7025140454?check_suite_focus=true

dranc:
Sorry I wasn't available to do the change earlier. Hope everything is ok now. :)

PR #249, opened 2022-07-07
fix MergeRequestCommentClient path

kagouball:
Regarding the doc (https://docs.gitlab.com/ee/api/notes.html#create-new-merge-request-note), the notes path and discussions path are preceded by "/projects/".


jairbubbles:
After checking with @kagouball we found out he was not using the class properly.

PR #257, opened 2022-07-26
Add possibility to get projects that have specific topics

louis-z:


NGitLab/Models/ProjectQuery.cs
```diff
@@ -64,6 +65,11 @@ public class ProjectQuery
         /// Limit results to projects with last_activity after specified time.
         /// </summary>
         public DateTimeOffset? LastActivityAfter;
+
+        /// <summary>
+        /// Limit results to projects that match all of the given topics.
+        /// </summary>
+        public List<string> Topics { get; } = new List<string>();
```
meziantou:
```suggestion
        public IList<string> Topics { get; } = new List<string>();
```

NGitLab/PublicAPI.Unshipped.txt
```diff
@@ -2433,6 +2433,7 @@ NGitLab.Models.ProjectQuery.Scope -> NGitLab.Models.ProjectQueryScope
 NGitLab.Models.ProjectQuery.Search -> string
 NGitLab.Models.ProjectQuery.Simple -> bool?
 NGitLab.Models.ProjectQuery.Statistics -> bool?
+NGitLab.Models.ProjectQuery.Topics.get -> System.Collections.Generic.List<string>
```
meziantou:
```suggestion
NGitLab.Models.ProjectQuery.Topics.get -> System.Collections.Generic.IList<string>
```

PR #259, opened 2022-08-01
Status in url should be lower case

kagouball:
In request url, pipeline status should be lower case : "....?status=...."

### DIFF
--- a/NGitLab/Impl/PipelineClient.cs
+++ b/NGitLab/Impl/PipelineClient.cs
@@ -145,7 +145,7 @@ namespace NGitLab.Impl
             if (query.Scope.HasValue)
                 queryEntries.Add("scope", query.Scope.Value.ToString());
             if (query.Status.HasValue)
-                queryEntries.Add("status", query.Status.Value.ToString());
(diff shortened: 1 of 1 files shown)


NGitLab/Impl/PipelineClient.cs
```diff
@@ -145,7 +145,7 @@ private string CreateSearchUrl(PipelineQuery query)
             if (query.Scope.HasValue)
                 queryEntries.Add("scope", query.Scope.Value.ToString());
             if (query.Status.HasValue)
-                queryEntries.Add("status", query.Status.Value.ToString());
+                queryEntries.Add("status", query.Status.Value.ToString().ToLower());
```
louis-z, line 148: As good practice, in most cases it's better to use `ToLowerInvariant()` than `ToLower()` (which is culture-sensitive).

meziantou: @louis-z We should add the BannedSymbol analyzer to ensure nobody uses this method (and a few other problematic methods)

louis-z: Good idea, I'll add it.

PR #260, opened 2022-08-01
Add Microsoft.CodeAnalysis.BannedApiAnalyzers + ban some symbols

louis-z:


BannedSymbols.txt
```diff
@@ -0,0 +1,3 @@
+﻿#https://github.com/dotnet/csharplang/blob/main/spec/documentation-comments.md#id-string-format
+
+M:System.String.ToLower;Use ToLowerInvariant instead
```
meziantou, line 3: I think you can add a few other common errors:

````
M:System.String.ToUpper;Use ToUpperInvariant instead

P:System.DateTime.Now;Use System.DateTime.UtcNow instead
P:System.DateTimeOffset.Now;Use System.DateTimeOffset.UtcNow instead
P:System.DateTimeOffset.DateTime;Use System.DateTimeOffset.UtcDateTime instead

F:System.StringComparison.CurrentCulture;Consider using Ordinal
F:System.StringComparison.CurrentCultureIgnoreCase;Consider using OrdinalIgnoreCase
F:System.StringComparison.InvariantCurrent;Consider using Ordinal
F:System.StringComparison.InvariantCurrentIgnoreCase;Consider using OrdinalIgnoreCase
````

PR #268, opened 2022-08-15
Indicate that project 'tag_list' is obsolete & redirect user to 'topics'

louis-z:


NGitLab.Mock/Clients/ProjectClient.cs
```diff
@@ -293,10 +293,17 @@ public Models.Project Update(string id, ProjectUpdate projectUpdate)
                     project.LfsEnabled = projectUpdate.LfsEnabled.Value;
                 }
 
+#pragma warning disable CS0618 // Type or member is obsolete
                 if (projectUpdate.TagList != null)
                 {
                     project.Tags = projectUpdate.TagList.Where(t => !string.IsNullOrEmpty(t)).Distinct(StringComparer.Ordinal).ToArray();
```
meziantou: Why not use topics here as both properties are equal?

PR #277, opened 2022-08-31
Adjust code so ProjectCreate.Tags and ProjectUpdate.TagList work again

louis-z:
Since https://github.com/ubisoft/NGitLab/pull/268 was merged, setting `ProjectCreate.Tags` or `ProjectUpdate.TagList` no longer had any effect, as the `Topics` property *(whose default value was an empty collection)* seemed to override data on the GitLab server side.

Make `Topics` null by default, in which case `Tags` or `TagList` can prevail.

### DIFF
--- a/NGitLab/Models/ProjectCreate.cs
+++ b/NGitLab/Models/ProjectCreate.cs
@@ -63,7 +63,7 @@ namespace NGitLab.Models
         public VisibilityLevel VisibilityLevel;
 
         [JsonPropertyName("tag_list")]
-        [Obsolete("Deprecated by GitLab. Use Topics instead")]
+        [Obsolete("Deprecated by GitLab. Use Topics instead", error: true)]
         public List<string> Tags;
 
         [JsonPropertyName("topics")]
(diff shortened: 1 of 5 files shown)


NGitLab/Models/ProjectCreate.cs
```diff
@@ -63,7 +63,7 @@ public class ProjectCreate
         public VisibilityLevel VisibilityLevel;
 
         [JsonPropertyName("tag_list")]
-        [Obsolete("Deprecated by GitLab. Use Topics instead")]
+        [Obsolete("Deprecated by GitLab. Use Topics instead", error: true)]
```
meziantou: I'm not sure we want to set `error: true`. The idea of `error:true` is to fail the build while providing a good error message. So, it should be for cases where the obsolete code doesn't work anymore.

This property is obsolete but still valid in GitLab, so users should be able to use it.

I think the actual fix would be to remove the initializer for the `Topics` property, so it can be null.

louis-z: I think that `error: true` is not only for cases where the obsolete code doesn't work anymore, but also when we truly want client code to stop using a member, but I see your point. Will revisit the fix.

meziantou: > but also when we truly want client code to stop using a member

AFAIK, the user cannot disable the error. So, they must update their code. This is not just a strong advise, this is mandatory.

Also, could you add a test to ensure ProjectUpdate doesn't override existing topics when the property is null.

PR #278, opened 2022-09-02
Adding support for group milestones and group issues

apoufortin:
Adding support for group milestones in the MilestoneClient.

### DIFF
--- a/NGitLab.Mock/Clients/GitLabClient.cs
+++ b/NGitLab.Mock/Clients/GitLabClient.cs
@@ -71,6 +71,8 @@
 
         public IMilestoneClient GetMilestone(int projectId) => new MilestoneClient(Context, projectId);
 
+        public IMilestoneClient GetGroupMilestone(int groupId) => new MilestoneClient(Context, groupId, "groups");
+
         public IReleaseClient GetReleases(int projectId) => new ReleaseClient(Context, projectId);
 
         public IPipelineClient GetPipelines(int projectId) => new PipelineClient(Context, jobClient: GetJobs(projectId), projectId: projectId);
--- a/NGitLab.Mock/Clients/GitLabClient.cs
+++ b/NGitLab.Mock/Clients/GitLabClient.cs
@@ -71,7 +71,7 @@
 
         public IMilestoneClient GetMilestone(int projectId) => new MilestoneClient(Context, projectId);
 
-        public IMilestoneClient GetGroupMilestone(int groupId) => new MilestoneClient(Context, groupId);
+        public IMilestoneClient GetGroupMilestone(int groupId) => throw new System.NotImplementedException();
 
         public IReleaseClient GetReleases(int projectId) => new ReleaseClient(Context, projectId);
 
--- a/NGitLab.Mock/Clients/GitLabClient.cs
+++ b/NGitLab.Mock/Clients/GitLabClient.cs
@@ -71,7 +71,7 @@
 
         public IMilestoneClient GetMilestone(int projectId) => new MilestoneClient(Context, projectId);
 
-        public IMilestoneClient GetGroupMilestone(int groupId) => new MilestoneClient(Context, groupId, "groups");
+        public IMilestoneClient GetGroupMilestone(int groupId) => new MilestoneClient(Context, groupId);
 
         public IReleaseClient GetReleases(int projectId) => new ReleaseClient(Context, projectId);
 
--- a/NGitLab/Impl/MilestoneClient.cs
+++ b/NGitLab/Impl/MilestoneClient.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using NGitLab.Extensions;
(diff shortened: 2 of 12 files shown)


NGitLab/Impl/MilestoneClient.cs
```diff
@@ -9,10 +9,10 @@ public class MilestoneClient : IMilestoneClient
         private readonly API _api;
         private readonly string _milestonePath;
 
-        public MilestoneClient(API api, int projectId)
+        public MilestoneClient(API api, int projectId, string urlPrefix = "projects")
```
louis-z: I don't like the idea of being able to create a `MilestoneClient` with just any `urlPrefix`...

Since a client application should always go through `IGitLabClient.GetMilestone()` or `IGitLabClient.GetGroupMilestone()` to create a `MilestoneClient`, I think we should instead introduce an internal ctor. Something like

```cs
internal MilestoneClient(API api, MilestoneScope scope, int id)
```

where `MilestoneScope` would be an enum that says whether we're interested in group or project milestones. `id` would then represent a group or project, accordingly.

louis-z (changes requested):

NGitLab/Impl/MilestoneClient.cs
```diff
@@ -6,13 +6,31 @@ namespace NGitLab.Impl
 {
     public class MilestoneClient : IMilestoneClient
     {
+        internal enum Scope
+        {
+            Project,
+            Milestone,
```
louis-z: Should be `Group`, right?

NGitLab.Mock/Clients/GitLabClient.cs
```diff
@@ -71,6 +71,8 @@ public IGraphQLClient GraphQL
 
         public IMilestoneClient GetMilestone(int projectId) => new MilestoneClient(Context, projectId);
 
+        public IMilestoneClient GetGroupMilestone(int groupId) => new MilestoneClient(Context, groupId);
```
louis-z: This calls the same ctor as GetMilestone. We need to distinguish between group and project scope.

NGitLab/Impl/MilestoneClient.cs
```diff
@@ -6,13 +6,31 @@ namespace NGitLab.Impl
...
+        }
+
         private readonly API _api;
         private readonly string _milestonePath;
 
-        public MilestoneClient(API api, int projectId)
```
louis-z, line 12: I would leave this ctor in for now, but tag it as obsolete. Basically, people should go through the GitLab client to instantiate a MilestoneClient.

louis-z (changes requested):

NGitLab/Impl/MilestoneClient.cs
louis-z: NVM, I saw it was moved further down. 🤦‍♂️

NGitLab/Impl/MilestoneClient.cs
apoufortin: crap!

NGitLab.Mock/Clients/GitLabClient.cs
apoufortin: for the test, i'll need to talk to you.

NGitLab/Impl/MilestoneClient.cs
```diff
@@ -6,13 +6,31 @@ namespace NGitLab.Impl
...
             _api = api;
-            _milestonePath = $"{Project.Url}/{projectId.ToStringInvariant()}/milestones";
+            string urlPrefix = "projects";
+
+            if (scope == Scope.Groups)
+            {
+                urlPrefix = "groups";
+            }
+
+            _milestonePath = $"/{urlPrefix}/{id.ToStringInvariant()}/milestones";
```
louis-z: You can simplify the code as follows:
```suggestion
            _milestonePath = $"/{scope.ToString().ToLowerInvariant()}/{id.ToStringInvariant()}/milestones";
```

NGitLab/Impl/MilestoneClient.cs
```diff
@@ -1,18 +1,31 @@
...
 
-        public MilestoneClient(API api, int projectId)
+        internal MilestoneClient(API api, Scope scope, int id)
         {
             _api = api;
-            _milestonePath = $"{Project.Url}/{projectId.ToStringInvariant()}/milestones";
+            _milestonePath = $"/{scope.ToString().ToLowerInvariant()}/{id.ToStringInvariant()}/milestones";
+        }
+
+        [Obsolete("Use GitlabClient.GetMilestone() or GitlabClient.GetGroupMilestone() instead.")]
```
louis-z: Make sure to use correct PascalCase:
```suggestion
        [Obsolete("Use GitLabClient.GetMilestone() or GitLabClient.GetGroupMilestone() instead.")]
```

louis-z:
Can you add 1 or more tests to `./NGitLab.Tests/Milestone/MilestoneClientTests.cs`, to validate creation and retrieval of group milestones?

PR #281, opened 2022-09-05
Add AvatarUrl to Assignee

mvarendorff:
This PR adds the missing optional AvatarUrl property to the Assignee class.

<sub>([API reference](https://docs.gitlab.com/ee/api/issues.html#single-issue))</sub>

### DIFF
--- a/NGitLab/Models/Assignee.cs
+++ b/NGitLab/Models/Assignee.cs
@@ -22,8 +22,8 @@ namespace NGitLab.Models
 
         [JsonPropertyName("created_at")]
         public DateTime CreatedAt;
-        
+
         [JsonPropertyName("avatar_url")]
-        public string? AvatarURL;
+        public string AvatarURL;
     }
 }
--- a/NGitLab/Models/Assignee.cs
+++ b/NGitLab/Models/Assignee.cs
@@ -22,5 +22,8 @@ namespace NGitLab.Models
 
         [JsonPropertyName("created_at")]
         public DateTime CreatedAt;
+        
+        [JsonPropertyName("avatar_url")]
+        public string? AvatarUrl;
     }
 }
--- a/NGitLab/Models/Assignee.cs
+++ b/NGitLab/Models/Assignee.cs
@@ -24,6 +24,6 @@ namespace NGitLab.Models
         public DateTime CreatedAt;
         
         [JsonPropertyName("avatar_url")]
-        public string? AvatarUrl;
+        public string? AvatarURL;
     }
 }
--- a/NGitLab.Mock/UserRef.cs
+++ b/NGitLab.Mock/UserRef.cs
@@ -61,6 +61,7 @@ namespace NGitLab.Mock
                 Email = Email,
                 Name = Name,
                 State = _user.State.ToString(),
+                AvatarURL = _user.AvatarUrl,
             };
         }
     }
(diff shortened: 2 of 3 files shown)


NGitLab/Models/Assignee.cs
```diff
@@ -22,5 +22,8 @@ public class Assignee
 
         [JsonPropertyName("created_at")]
         public DateTime CreatedAt;
+        
+        [JsonPropertyName("avatar_url")]
+        public string? AvatarURL;
```
jairbubbles: If you wish to use nullable you must add `#nullable enable` at the beginning of the file. Not adding nullability is fine too.

Also, you need to add it to the public API, there should be a quick fix in your IDE to do so easily.

jairbubbles (commented):
Thx for the PR @geisterfurz007!

Could you also update the usage of Assignee, notably `ToClientAssignee` in `UserRef.cs`

NGitLab/Models/Assignee.cs
mvarendorff: Ah, I should have known it wasn't as easy as blindly editing a file in the GitHub editor, my apologies 😄 I *think* I caught everything you mentioned; if there is anything missing, let me know!

jairbubbles: You mean you didn't test at all? 😱

mvarendorff: That's correct; I didn't think much of it, when I spotted that the field was missing when replacing GitLabApiClient with this in one of my work projects. I just wanted to raise a PR since I figured it'd just be a quick edit.

jairbubbles: Ok, I checked quickly and I feel like  there are other fields which are not correct:

```
 "assignee" : {
         "state" : "active",
         "id" : 1,
         "name" : "Administrator",
         "web_url" : "https://gitlab.example.com/root",
         "avatar_url" : null,
         "username" : "root"
      }
```
(see https://docs.gitlab.com/ee/api/issues.html#single-issue)

@meziantou @louis-z  Should we fix all the fields? The struct is very old and doesn't match the current documentation.

louis-z: Certain fields seem indeed to be missing from the GitLab response: `email`, `created_at`. Maybe we should plan a wider cleanup, to remove any irrelevant fields throughout the code base and - we can always dream - convert public fields to properties.

jairbubbles: What is tricky is that sometimes some DTO classes are used in multiple routes that doesn't always return the same fields. But it doesn't seem to be the case here.

meziantou: Some fields are only accessible for admin users. This is common for email addresses. Maybe this need more tests before deleting fields.

jairbubbles:
Thx @geisterfurz007  and sorry for the delay.

mvarendorff:
No worries @jairbubbles! I appreciate the guidance :) Quick follow-up question though: I noticed there hadn't been a new release and saw that my commit was causing a bunch of failed tests; the logs look like it may not have been my fault but I am not sure. If I broke something there, let me know and I will get to fixing that asap, of course!

jairbubbles:
Sadly the CI is a bit flaky but don't worry it's probably not related to your changes. I'll release a new version right away.

jairbubbles:
https://github.com/ubisoft/NGitLab/releases/tag/6.27.3 (it should take around 20 min)

mvarendorff:
Awesome, thanks a bunch! I appreciate it a lot :)

PR #285, opened 2022-09-07
The mocks now handle accepting MRs from forks

louis-z:
Mocks did not correctly accept MRs from one project to another.

### DIFF
--- a/NGitLab.Mock/Repository.cs
+++ b/NGitLab.Mock/Repository.cs
@@ -200,7 +200,9 @@ namespace NGitLab.Mock
         public void RemoveBranch(string branchName)
         {
             var repository = GetGitRepository();
-            repository.Branches.Remove(branchName);
+            Commands.Checkout(repository, Project.DefaultBranch);
+            var branch = repository.Branches[branchName];
+            repository.Branches.Remove(branch);
         }
 
         public TagCollection GetTags()
--- a/NGitLab.Mock/Repository.cs
+++ b/NGitLab.Mock/Repository.cs
@@ -473,6 +473,22 @@ namespace NGitLab.Mock
             }
         }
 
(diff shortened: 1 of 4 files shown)


NGitLab.Mock/Repository.cs
```diff
@@ -473,6 +475,22 @@ internal IEnumerable<Models.Tree> GetTree(Models.RepositoryGetTreeOptions reposi
             }
         }
 
+        internal string FetchBranchFromFork(Project fork, string branch)
+        {
+            if (fork?.ForkedFrom != Project)
+                throw new InvalidOperationException("Cannot fetch MR source branch from unrelated project");
+
+            var repo = GetGitRepository();
+            var forkRemoteName = "fork"; // Should we also include the name of the fork?
```
meziantou: I think it should be unique per fork. If you have multiple operation in parallel, this is not safe.

Clients are safe, because all code run in a lock. However, you can manipulate the mock objects directly, and there is no thread-safe guarantee.

```suggestion
            var forkRemoteName = "fork" + fork.ForkedFrom.Id;
```

NGitLab.Mock/Repository.cs
```diff
@@ -473,6 +475,22 @@ internal IEnumerable<Models.Tree> GetTree(Models.RepositoryGetTreeOptions reposi
...
+            var forkRemote =
+                repo.Network.Remotes[forkRemoteName] ??
+                repo.Network.Remotes.Add(forkRemoteName, fork.Repository.FullPath);
+
+            Commands.Fetch(repo, forkRemote.Name, new[] { branch }, new FetchOptions { }, "");
```
meziantou: Could you use named arguments for the last argument (`""`)?

PR #288, opened 2022-09-09
Improve MR & Project mock logic

louis-z:


NGitLab.Mock/Project.cs
```diff
@@ -296,6 +297,57 @@ public MergeRequest CreateNewMergeRequest(User user, string branchName, string t
...
+            {
+                consolidatedSourceBranchName = targetProject.Repository.FetchBranchFromFork(sourceProject, sourceBranchName);
+            }
+
+            var commonCommit = targetProject.Repository.FindMergeBase(targetBranchCommit, sourceBranchCommit);
+            if (commonCommit is null)
+                throw new InvalidOperationException($"Branch '{sourceBranchName}' does not seem to stem from branch '{targetBranchName}'");
+
+            // Determine if there are conflicts, by performing a rebase on a volatile copy of the source branch
+            var volatileBranchName = "volatile-" + Guid.NewGuid().ToString("N");
```
meziantou:
```suggestion
            var volatileBranchName = "volatile/" + Guid.NewGuid().ToString("N");
```

NGitLab.Mock/Project.cs
```diff
@@ -296,6 +297,57 @@ public MergeRequest CreateNewMergeRequest(User user, string branchName, string t
...
+            targetProject.Repository.CreateBranch(volatileBranchName, consolidatedSourceBranchName);
+            var hasConflicts = !targetProject.Repository.Rebase(user, volatileBranchName, targetBranchName);
```
meziantou: I don't think this should be checked once. Indeed, the value is obsolete when the source or target branch is updated.

I think this should be a computed property or a method

louis-z: You're right. I made the code somewhat better, recomputing the MR's internal state whenever necessary.

PR #301, opened 2022-09-21
Correctly set ProjectId in Pipeline mock

DomLatr:


NGitLab.Mock/Pipeline.cs
```diff
@@ -18,7 +18,7 @@ public Pipeline(string @ref)
 
         public int Id { get; set; }
 
-        public int ProjectId { get; set; }
+        public int ProjectId => Project.Id;
```
louis-z: If someone creates a `Pipeline` object manually, the `Project` property may be null
```suggestion
        public int ProjectId => Project?.Id ?? 0;
```

louis-z (changes requested):

PR #303, opened 2022-09-22
PipelineClient supports Test Reports but not Test Reports Summaries

### LINKED ISSUE #302: PipelineClient supports Test Reports but not Test Reports Summaries
The current `PipelineClient` can get the test reports using [this endpoint](https://docs.gitlab.com/ee/api/pipelines.html#get-a-pipelines-test-report),  
```csharp
public TestReport GetTestReports(int pipelineId)
{
    return _api.Get().To<TestReport>($"{_projectPath}/pipelines/{pipelineId.ToStringInvariant()}/test_report");
}
```

but not the [summaries](https://docs.gitlab.com/ee/api/pipelines.html#get-a-pipelines-test-report-summary), which would be the faster call, if the full details are not needed.

msrouchou:
- Closes #302

### DIFF
--- a/NGitLab.Mock.Tests/PipelineTests.cs
+++ b/NGitLab.Mock.Tests/PipelineTests.cs
@@ -21,18 +21,27 @@ namespace NGitLab.Mock.Tests
 
             var reportSummary = pipeline.TestReportsSummary = new Models.TestReportSummary
             {
-                TotalTime = 60,
-                TotalCount = 1157,
-                SuccessCount = 1157,
+                Total = new Models.TestReportSummaryTotals
+                {
+                    Time = 60,
+                    Count = 1157,
+                    Success = 1157,
+                    Failed = 0,
+                    Skipped = 0,
+                    Error = 0,
+                },
             };
 
             var client = server.CreateClient();
             Assert.AreEqual(job.Trace, await client.GetJobs(project.Id).GetTraceAsync(job.Id));
 
             var summary = client.GetPipelines(project.Id).GetTestReportsSummary(pipeline.Id);
-            Assert.AreEqual(60, summary.TotalTime);
-            Assert.AreEqual(1157, summary.TotalCount);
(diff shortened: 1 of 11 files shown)


NGitLab/Models/TestReportSummary.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Runtime.Serialization;
+
+namespace NGitLab.Models
+{
+    [DataContract(Name = "total")]
```
Toa741: I don't think that will work.

`TestReport` and `TestReportSummary` does share 'test_suites', but other props are not the same.

If you are doing that, I would bet that you will have the 'total' property empty (cause of different names) and no 'test_suites' (cause it will not find any under 'total' prop)

I don't think `TestReportSummary` will be able to share much with `TestReport` except the `TestSuites` prop.

meziantou: If the 2 methods return different things, then create 2 different classes.

msrouchou: Fixed.

NGitLab.Mock.Tests/PipelineTests.cs
```diff
@@ -15,11 +15,33 @@ public async Task Test_pipelines()
             var commit = project.Repository.Commit(user, "test");
 
             var pipeline = project.Pipelines.Add(commit.Sha, JobStatus.Success, user);
+
             var job = pipeline.AddNewJob("Test_job", JobStatus.Success);
             job.Trace = "This is a trace\nWith Multiple line";
 
+            var reportSummary = pipeline.TestReportsSummary = new Models.TestReportSummary
```
Toa741: In this case, can it have its own test? That would discourage adding more in this Test_pipelines test

msrouchou: seperate test created :)

PR #315, opened 2022-10-06
Add epic_id and epic_iid properties

tbavard:
Add epic_id and epic_iid properties so we can Edit issue and assigned an Epic

### DIFF
--- a/NGitLab/Models/IssueCreate.cs
+++ b/NGitLab/Models/IssueCreate.cs
@@ -39,6 +39,5 @@ namespace NGitLab.Models
 
         [JsonPropertyName("epic_iid")]
         public int? EpicIId;
-
     }
 }
--- a/NGitLab/Models/IssueCreate.cs
+++ b/NGitLab/Models/IssueCreate.cs
@@ -36,8 +36,5 @@ namespace NGitLab.Models
 
         [JsonPropertyName("epic_id")]
         public int? EpicId;
-
-        [JsonPropertyName("epic_iid")]
-        public int? EpicIid;
     }
 }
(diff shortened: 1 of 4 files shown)


NGitLab/Models/IssueCreate.cs
```diff
@@ -33,5 +33,11 @@ public class IssueCreate
         [JsonPropertyName("due_date")]
         [JsonConverter(typeof(DateOnlyConverter))]
         public DateTime? DueDate;
+
+        [JsonPropertyName("epic_id")]
+        public int? EpicId;
+
+        [JsonPropertyName("epic_iid")]
+        public int? EpicIId;
```
louis-z: Don't add this field, as it's been deprecated by GitLab and will eventually be removed. See [here](https://docs.gitlab.com/ee/api/issues.html#new-issue).

NGitLab/Models/IssueEdit.cs
```diff
@@ -36,5 +36,11 @@ public class IssueEdit
         [JsonPropertyName("due_date")]
         [JsonConverter(typeof(DateOnlyConverter))]
         public DateTime? DueDate;
+
+        [JsonPropertyName("epic_id")]
+        public int? EpicId;
+
+        [JsonPropertyName("epic_iid")]
+        public int? EpicIId;
```
louis-z: `epic_iid` is deprecated and scheduled for removal.

louis-z (changes requested):

NGitLab/Models/IssueCreate.cs
tbavard: yes I know, but we need this field in older gitlab version like we used at Ubisoft. Epic_id needs extra permission

PR #318, opened 2022-10-11
Improve testing for Users client

### LINKED ISSUE #317: Improve testing for Users client
- Add a test to search for user by email. GitLab changed its behavior and recent versions require an admin client.
- Update the GitLab versions we test the library against

msrouchou:
- Closes #317

### DIFF
--- a/NGitLab.Tests/UsersTests.cs
+++ b/NGitLab.Tests/UsersTests.cs
@@ -200,6 +200,7 @@ namespace NGitLab.Tests
             {
                 Email = $"test{randomNumber}@test.pl",
                 PublicEmail = $"test{randomNumber}@gmail.com",
+                SkipConfirmation = true, // Skip confirmation and assume public email is verified
                 Bio = "bio",
                 CanCreateGroup = true,
                 IsAdmin = true,
--- a/NGitLab.Tests/UsersTests.cs
+++ b/NGitLab.Tests/UsersTests.cs
@@ -54,6 +54,28 @@ namespace NGitLab.Tests
 
         [Test]
         [NGitLabRetry]
+        public async Task GetUserByEmailDoesNotWorkOnNonAdminClient()
+        {
+            using var context = await GitLabTestContext.CreateAsync();
+            var addedUser = CreateNewUser(context);
+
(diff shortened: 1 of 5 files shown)


NGitLab.Tests/UsersTests.cs
```diff
@@ -52,6 +52,28 @@ public async Task CreateUpdateDelete()
             await GitLabTestContext.RetryUntilAsync(() => users.Get(addedUser.Username).ToList(), users => !users.Any(), TimeSpan.FromMinutes(2));
         }
 
+        [Test]
+        [NGitLabRetry]
+        public async Task GetUserByEmailDoesNotWorkOnNonAdminClient()
```
louis-z, line 57: Can you add a test that searches a user using their `public_email` instead? I believe that should work without requiring admin rights.

louis-z (changes requested):

NGitLab.Tests/UsersTests.cs
meziantou: There should be a test to validate
- admin can search by private email
- non-admin cannot search by private email
- non-admin can search by public email

msrouchou: Unfortuenately, it was not possible to test against the public_email:
- According the the API, we could provide a public_email with [Update (Put /users/:id)](https://docs.gitlab.com/ee/api/users.html#user-modification) but [the update itself fails because of email verification](https://github.com/ubisoft/NGitLab/runs/8849941598#r0s46): 
`NGitLab.GitLabException : GitLab server returned an error (BadRequest): {"public_email":["must be an email you have verified"]}. Original call: Put http://localhost:48624/api/v4/users/20. With data {"public_email":"test1499957960@gmail.com","skip_confirmation":true,"skip_reconfirmation":true}`

louis-z: Ok then... since it's not as easy as one might have thought initially, let's forget about that test for now.

msrouchou: Discussed with Louis and we decided it is ok if we don't test the public email as it is not doable via the API alone.

PR #320, opened 2022-10-14
Resourcelabelevent mock

phdesroUbi:


NGitLab.Mock/Clients/IssueClient.cs
```diff
@@ -85,12 +85,25 @@ public Models.Issue Edit(IssueEdit issueEdit)
...
+                {
+                    labelsEdit = Array.Empty<string>();
+                }
+
+                if (labelsEdit is not null)
+                {
+                    CreateResourceLabelEvents(issueToModify.Labels, labelsEdit, issueToModify.Id);
+                }
+
+                issueToModify.Labels = labelsEdit ?? issueToModify.Labels;
```
louis-z: I think you can simplify the logic to something like
```suggestion
                if (!string.IsNullOrWhiteSpace(issueEdit.Labels))
                {
                    var labelsEdit = issueEdit.Labels.Split(',');
                    CreateResourceLabelEvents(issueToModify.Labels, labelsEdit, issueToModify.Id);
                    issueToModify.Labels = labelsEdit;
                }
```

NGitLab.Mock/Clients/IssueClient.cs
```diff
@@ -361,5 +384,64 @@ private IEnumerable<Issue> FilterByQuery(IEnumerable<Issue> issues, IssueQuery q
 
             return issues;
         }
+
+        public void CreateResourceLabelEvents(string[] previousLabels, string[] newLabels, int resourceId)
```
louis-z: This method can be private.

louis-z (changes requested):

NGitLab.Mock/Clients/IssueClient.cs
phdesroUbi: It's not the same behavior if it's null or if it's an empty string.

louis-z: Ok, but I'm sure you can still simplify and/or make the logic easier to understand.

louis-z (approved):

PR #325, opened 2022-10-21
Introduce QueryParameterAttribute & QueryStringHelper

louis-z:
Eliminate query string boilerplate by introducing the QueryParameterAttribute and the QueryStringHelper.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -8,7 +8,7 @@
     <Authors>$(Company), NGitLab contributors</Authors>
 
     <!-- Compile Options -->
-    <LangVersion>9.0</LangVersion>
+    <LangVersion>10.0</LangVersion>
     <Deterministic>true</Deterministic>
     <Features>strict</Features>
     <TreatWarningsAsErrors Condition="'$(Configuration)' != 'Debug'">true</TreatWarningsAsErrors>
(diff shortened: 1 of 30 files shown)


meziantou:
I don't know if we want to accept this change.
- It doesn't really change the number of line of code, it just moves it (1 direct call to `Utils.AddParameter` vs 1 attribute)
- We don't add lots of new code, so it doesn't really solve an existing problem
- It uses reflection, so not AOT friendly, but the rest of the code is not AOT friendly too at the moment (JSON serializer)
- I don't want to review the full PR that migrate all existing code to the new way 😝

I'll let other maintainers give give their opinion on this change. @Toa741

louis-z:
For now, it's just a proof of concept, but I feel it might help developers in the long run by
- Making the handling of query parameters similar to that of JSON objects (albeit using different attributes)
- Favoring separation of concerns (i.e. `UserQuery` may be better suited than `UserClient` to determine what data should be present in the query string and how it should be formatted)

And of course, if we were to choose this route, we would migrate the code gradually, so nobody would have to review one big PR...

Pfvc298:
I'm divided as well.
- I like that it makes things simpler
- Some cases might not be covered (I'm pretty sure one of the endpoint of environments accepts array of strings for ex), that'll work also up to the point we find other edge cases

Toa741:
I'm divided as well.

I like that it makes things simpler
Some cases might not be covered (I'm pretty sure one of the endpoint of environments accepts array of strings for ex), that'll work also up to the point we find other edge cases

PR #328, opened 2022-10-28
Add ReleaseInfoLinks

louis-z:


NGitLab.Tests/ReleaseClientTests.cs
```diff
@@ -33,6 +33,8 @@ public async Task Test_release_api()
             Assert.That(release.TagName, Is.EqualTo("0.7"));
             Assert.That(release.Name, Is.EqualTo("0.7"));
             Assert.That(release.Description, Is.EqualTo("test"));
+            Assert.That(release.Links.Self, Is.EqualTo($"{context.HttpClient.BaseAddress}{project.PathWithNamespace}/-/releases/0.7"));
```
meziantou: I think you can use `GitLabServer.Url`

louis-z: I used `project.WebUrl` instead

PR #346, opened 2022-12-09
Add ITagClient.GetByNameAsync

louis-z:
Add method to retrieve a specific git tag


NGitLab/ITagClient.cs
```diff
@@ -7,6 +7,8 @@ public interface ITagClient
     {
         Tag Create(TagCreate tag);
 
+        Tag Get(string name);
```
meziantou: I think we should add async methods when we map new endpoint. 

If you want you can keep both methods, but the async one seems the most important.

PR #354, opened 2022-12-20
Add IMergeRequestClient.GetByIidAsync

meziantou:


NGitLab/Models/MergeRequestGet.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System.Text.Json.Serialization;
+
+namespace NGitLab.Models
+{
+    public class MergeRequestGet
```
Toa741: Most of the class for Get have "Query" as a suffix, keep it for consistency?

meziantou: I though `Query` was for methods that return a collection, but I found other example for methods that return a single element. The code is updated.

PR #370, opened 2023-01-17
Path of Repositories ends with namespace and name

baudronp:


NGitLab.Mock/Repository.cs
```diff
@@ -67,9 +68,11 @@ private LibGit2Sharp.Repository GetGitRepository()
                     if (_directory == null)
                     {
                         var directory = TemporaryDirectory.Create();
+                        var repositoryDirectory = Path.Combine(directory.FullPath, Project.PathWithNamespace.Replace('/', Path.DirectorySeparatorChar));
```
meziantou, line 73: Can you add a comment to explain why having the namespace in the path is important

global.json
```diff
@@ -1,6 +1,6 @@
 {
     "sdk": {
         "version": "6.0.202",
-        "rollForward": "latestFeature"
+        "rollForward": "major"
```
louis-z, line 4: Not sure we want this.

NGitLab.Mock.Tests/ProjectsMockTests.cs
```diff
@@ -35,6 +35,23 @@ public void Test_project_can_be_cloned_by_default()
...
+                .WithUser("Test", isDefault: true)
+                .WithProject("Test", @namespace: "testgroup", addDefaultUserAsMaintainer: true)
+                .BuildServer();
+
+            var client = server.CreateClient();
+            var project = client.Projects["testgroup/Test"];
+
+            Assert.IsNotNull(project);
+            StringAssert.EndsWith("testgroup\\test", project.SshUrl);
+            StringAssert.EndsWith("testgroup\\test", project.HttpUrl);
```
louis-z: Should be slashes, not backslashes, right?

global.json
meziantou: Most people don't have a compatible SDK on their machine. Using `major`, you can open and build the solution correctly.

NGitLab.Mock.Tests/ProjectsMockTests.cs
baudronp: They are "files://" uri with '\' in them.

baudronp: in fact this seems to be platform dependant... I fix it shortly

PR #380, opened 2023-02-03
Make sure transient branches are always removed

jairbubbles:


NGitLab.Mock/MergeRequest.cs
```diff
@@ -382,7 +382,14 @@ private void RefreshInternalState()
             // Determine if there are conflicts, by performing a rebase on a transient copy of the source branch
             var transientBranchName = "transient/" + Guid.NewGuid().ToString("N");
             Project.Repository.CreateBranch(transientBranchName, _consolidatedSourceBranch);
-            _hasConflicts = !Project.Repository.Rebase(DefaultUser, transientBranchName, TargetBranch);
+            try
+            {
+                _hasConflicts = !Project.Repository.Rebase(DefaultUser, transientBranchName, TargetBranch);
```
meziantou, line 387: If rebase return a `bool`, I expect this method to catch exception.

jairbubbles: It's a blind fix, in some UI tests video we've been seeing some "transient" branches pop up.

louis-z: I agree with @meziantou, but catching exceptions in the `Rebase` method and returning false instead might impact other call sites, especially those that ignore the return value.

Since we seem to need a quick fix, let's merge this and revisit the code in another PR.

meziantou: @louis-z Should we create the transient branch in a _custom_ ref path such as `refs/rebase/{unique name}`? I think most git tools would ignore these refs when listing branches, so it would fix your issue.

louis-z: It's worth trying out.

PR #383, opened 2023-02-10
Make transient branch not fetchable for cloned project

LePtitDev:
Based on https://github.com/ubisoft/NGitLab/pull/380#discussion_r1096012621

### DIFF
--- a/NGitLab.Mock/MergeRequest.cs
+++ b/NGitLab.Mock/MergeRequest.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
+using System.IO;
 using System.Linq;
 using System.Security.Cryptography;
 using System.Text;
@@ -380,8 +381,13 @@ namespace NGitLab.Mock
             _divergedCommitsCount = Project.Repository.ComputeDivergence(TargetBranchHeadCommit, commonCommit);
 
             // Determine if there are conflicts, by performing a rebase on a transient copy of the source branch
-            var transientBranchName = "transient/" + Guid.NewGuid().ToString("N");
-            Project.Repository.CreateBranch(transientBranchName, _consolidatedSourceBranch);
+            var smallBranchName = Guid.NewGuid().ToString("N");
+            var transientBranchName = "refs/transient/" + smallBranchName;
+
+            var branchFullPath = Path.Combine(Project.Repository.FullPath, ".git/refs/transient", smallBranchName);
+            Directory.CreateDirectory(Path.GetDirectoryName(branchFullPath)!);
+            System.IO.File.WriteAllText(branchFullPath, Project.Repository.GetCommit(SourceBranch).Sha);
+
             try
             {
                 _hasConflicts = !Project.Repository.Rebase(DefaultUser, transientBranchName, TargetBranch);
@@ -391,7 +397,7 @@ namespace NGitLab.Mock
                 _hasConflicts = true;
             }
 
-            Project.Repository.RemoveBranch(transientBranchName);
+            //System.IO.File.Delete(branchFullPath);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


NGitLab.Mock/MergeRequest.cs
```diff
@@ -380,8 +381,13 @@ private void RefreshInternalState()
             _divergedCommitsCount = Project.Repository.ComputeDivergence(TargetBranchHeadCommit, commonCommit);
 
             // Determine if there are conflicts, by performing a rebase on a transient copy of the source branch
-            var transientBranchName = "transient/" + Guid.NewGuid().ToString("N");
-            Project.Repository.CreateBranch(transientBranchName, _consolidatedSourceBranch);
+            var smallBranchName = Guid.NewGuid().ToString("N");
+            var transientBranchName = "refs/transient/" + smallBranchName;
+
+            var branchFullPath = Path.Combine(Project.Repository.FullPath, ".git/refs/transient", smallBranchName);
```
meziantou: We are using Libgit2sharp in the repo. So, I think you can use `repo.Refs.Add(transientBranchName, ...)`. Make you can need to add an internal method in `Project.Repository` or expose a way to get the LibGit2Sharp repository object.

LePtitDev: I can add a ref with LibGit2Sharp but this ref is not usable for rebase. I tried an other way, I use the merge without committing to check conflicts and I abort operation after that. See : https://stackoverflow.com/a/501461

NGitLab.Mock/Repository.cs
```diff
@@ -636,6 +636,43 @@ public bool Rebase(UserRef user, string sourceBranch, string targetBranch)
...
+            var signature = new Signature(user.UserName, user.Email, DateTimeOffset.UtcNow);
+            var options = new MergeOptions
+            {
+                CommitOnSuccess = false,
+                FastForwardStrategy = FastForwardStrategy.NoFastForward,
+            };
+
+            Commands.Checkout(repo, upstream);
+
+            var resetTip = upstream.Tip;
```
meziantou, line 663: Why is this needed? As you don't create a commit, the `Tip` should not change?

NGitLab.Mock/Repository.cs
```diff
@@ -636,6 +636,43 @@ public bool Rebase(UserRef user, string sourceBranch, string targetBranch)
...
+            if (string.IsNullOrEmpty(targetBranch))
+                throw new ArgumentException("Cannot be null or empty", nameof(targetBranch));
+
+            var repo = GetGitRepository();
+
+            var head = repo.Head;
+            var branch = repo.Branches[sourceBranch];
+            var upstream = repo.Branches[targetBranch];
+
+            _ = repo.Network.Remotes[targetBranch];
```
meziantou: Is it needed?

NGitLab.Mock/Repository.cs
```diff
@@ -636,6 +636,43 @@ public bool Rebase(UserRef user, string sourceBranch, string targetBranch)
...
+
+            Commands.Checkout(repo, upstream);
+
+            var resetTip = upstream.Tip;
+
+            var mergeResult = repo.Merge(branch.Tip, signature, options);
+            var result = mergeResult.Status == MergeStatus.Conflicts;
+
+            repo.Reset(ResetMode.Hard, resetTip);
+            Commands.Checkout(repo, head);
```
meziantou, line 669: Do you think we can use `new CheckoutOptions { CheckoutModifiers = CheckoutModifiers.Force }` and remove the previous `Reset`? (@louis-z)

louis-z: That sounds reasonable. It's worth a shot.

LePtitDev: If I don't use `Reset`, the merging operation is not aborted

image

meziantou: You are right

PR #387, opened 2023-02-13
Implement subgroups endpoint

### LINKED ISSUE #295: Subgroups api missing or not readily visible
I've been trying to get the subgroups for a group, however I don't see any other way than to get all groups and then filter by parent id. This seems a bit roundabout, seeing how gitlab does offer a list subgroup endpoint:

https://docs.gitlab.com/ee/api/groups.html#list-a-groups-subgroups

This endpoint is basically a scoped groups endpoint.
This plus the list descendant groups would be nice additions. We heavily structure our projects in subgroups within subgroups so having something to work with that would be nice.

Rimgelair:
Fix https://github.com/ubisoft/NGitLab/issues/295

### DIFF
--- a/NGitLab/IGroupsClient.cs
+++ b/NGitLab/IGroupsClient.cs
@@ -36,6 +36,10 @@ namespace NGitLab
 
         Task<Group> GetByFullPathAsync(string fullPath, CancellationToken cancellationToken = default);
 
+        GitLabCollectionResponse<Group> GetSubgroupsByIdAsync(int id, GroupQuery query);
+
+        GitLabCollectionResponse<Group> GetSubgroupsByFullPathAsync(string fullPath, GroupQuery query);
+
         Group Create(GroupCreate group);
 
         Task<Group> CreateAsync(GroupCreate group, CancellationToken cancellationToken = default);
--- a/NGitLab/IGroupsClient.cs
+++ b/NGitLab/IGroupsClient.cs
@@ -36,9 +36,9 @@ namespace NGitLab
 
         Task<Group> GetByFullPathAsync(string fullPath, CancellationToken cancellationToken = default);
 
-        GitLabCollectionResponse<Group> GetSubgroupsByIdAsync(int id, GroupQuery query);
+        GitLabCollectionResponse<Group> GetSubgroupsByIdAsync(int id, SubgroupQuery query = null);
 
-        GitLabCollectionResponse<Group> GetSubgroupsByFullPathAsync(string fullPath, GroupQuery query);
+        GitLabCollectionResponse<Group> GetSubgroupsByFullPathAsync(string fullPath, SubgroupQuery query = null);
 
         Group Create(GroupCreate group);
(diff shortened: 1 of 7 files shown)


NGitLab/Impl/GroupsClient.cs
```diff
@@ -112,6 +117,18 @@ public Task<Group> GetByFullPathAsync(string fullPath, CancellationToken cancell
             return _api.Get().ToAsync<Group>(Url + "/" + Uri.EscapeDataString(fullPath), cancellationToken);
         }
 
+        public GitLabCollectionResponse<Group> GetSubgroupsByIdAsync(int id, GroupQuery query)
```
meziantou: I think query should allow null and be optional. Something like 

```suggestion
        public GitLabCollectionResponse<Group> GetSubgroupsByIdAsync(int id, GroupQuery query = null)
```

NGitLab/IGroupsClient.cs
```diff
@@ -36,6 +36,10 @@ public interface IGroupsClient
 
         Task<Group> GetByFullPathAsync(string fullPath, CancellationToken cancellationToken = default);
 
+        GitLabCollectionResponse<Group> GetSubgroupsByIdAsync(int id, GroupQuery query);
```
meziantou: You can't reuse `GroupQuery` here. The available options are not the same (e.g. `top_level_only`). You must create new class `SubgroupQuery`.

meziantou (changes requested):

NGitLab/Impl/GroupsClient.cs
Rimgelair: I agree I have made the changes.

NGitLab/IGroupsClient.cs
Rimgelair: That was an oversight haha, I only checked the GroupQuery class when I should have cross referenced with he docs. I have made the change.

NGitLab/Models/SubgroupQuery.cs
```diff
@@ -0,0 +1,63 @@
...
+    /// Allows to use more advanced GitLab queries for getting subgroups (based on v4 GitLab API).
+    /// https://docs.gitlab.com/ee/api/groups.html#list-a-groups-subgroups
+    /// </summary>
+    public class SubgroupQuery
+    {
+        /// <summary>
+        /// Skip the group IDs passed
+        /// (optional)
+        /// </summary>
+        public int[] SkipGroups;
```
meziantou: Use properties everywhere (using fields is _legacy_)

```suggestion
        public int[] SkipGroups { get; set; }
```

meziantou (changes requested):

NGitLab/Models/SubgroupQuery.cs
Rimgelair: Done 😄

meziantou (approved):

PR #393, opened 2023-02-26
Add GetVersions to the merge request client.

waaverecords:


NGitLab/IMergeRequestClient.cs
```diff
@@ -45,6 +45,8 @@ public interface IMergeRequestClient
 
         IEnumerable<Author> GetParticipants(int mergeRequestIid);
 
+        IEnumerable<MergeRequestVersion> GetVersions(int mergeRequestIid);
```
meziantou: We try expose `async` methods when mapping a new methods. In this case, only the return type need to be changed.

```suggestion
        GitLabCollectionResponse<MergeRequestVersion> GetVersionsAsync(int mergeRequestIid);
```

meziantou (changes requested):
Thanks for the contribution!

Do you think you can add a test for this change? (In this file: https://github.com/ubisoft/NGitLab/blob/main/NGitLab.Tests/MergeRequest/MergeRequestClientTests.cs)

NGitLab/IMergeRequestClient.cs
waaverecords: 👍

meziantou (approved):

PR #403, opened 2023-03-07
Use Microsoft.Playwright to set up Docker test container

louis-z:
Plus, remove dependency on AngleSharp

### DIFF
--- a/NGitLab.Tests/Docker/GitLabDockerContainer.cs
+++ b/NGitLab.Tests/Docker/GitLabDockerContainer.cs
@@ -12,12 +12,9 @@ using System.Runtime.InteropServices;
 using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
-using AngleSharp;
-using AngleSharp.Dom;
-using AngleSharp.Html.Dom;
-using AngleSharp.Io;
 using Docker.DotNet;
 using Docker.DotNet.Models;
+using Microsoft.Playwright;
 using NGitLab.Models;
 using NUnit.Framework;
 using Polly;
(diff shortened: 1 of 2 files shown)


NGitLab.Tests/Docker/GitLabDockerContainer.cs
```diff
@@ -278,127 +275,94 @@ private async Task GenerateCredentialsAsync()
...
-                    result = await SignIn(context);
+                    await page.Locator("form#new_user input[name='user[login]']").FillAsync(AdminUserName);
+                    await page.Locator("form#new_user input[name='user[password]']").FillAsync(AdminPassword);
+
+                    var checkbox = page.Locator("form#new_user input[type=checkbox][name='user[remember_me]']");
+                    await checkbox.CheckAsync(new LocatorCheckOptions { Force = true });
+
+                    await page.RunAndWaitForResponseAsync(async () =>
+                    {
+                        await page.EvalOnSelectorAsync("form#new_user", "form => form.submit()");
```
Toa741, line 308: Could it simply be replaced with a click on the form submit button?

NGitLab.Tests/Docker/GitLabDockerContainer.cs
```diff
@@ -278,127 +275,94 @@ private async Task GenerateCredentialsAsync()
...
+                    }, GitLabUrl.AbsoluteUri);
+
+                    url = await GetCurrentUrl(page);
```
Toa741, line 311: page.Url doesn't work here?

NGitLab.Tests/Docker/GitLabDockerContainer.cs
meziantou: Clicking on a button means more maintenance. If the UI change => KO.

NGitLab.Tests/Docker/GitLabDockerContainer.cs
louis-z: It partially works... but sometimes with a delay. For instance, say you make these 3 calls in a row

```cs
url = page.Url;
url = await GetCurrentUrl(page);
url = page.Url;
```

- the 1st time around, all 3 calls will return the same thing: `/users/sign_in`
- the 2nd time around, the 1st call will return `/users/sign_in`, but the 2nd and 3rd calls will return `/`

PR #407, opened 2023-03-14
Cleanup related to project ID unnecessarily expressed as JSON field

louis-z:
- Remove `[JsonPropertyName("id")]` from project ID fields/properties needlessly serialized as JSON and passed to `POST` or `PUT` commands.
- In cases where those fields/properties truly are unnecessary, tag as `[EditorBrowsable(EditorBrowsableState.Never)]` _(they might eventually become `[Obsolete]`)_.
- Delete `ApprovalRuleDelete` class, never used
- Running tests, we see that `CommitStatus.ProjectId` _(= 101, here below)_ does NOT correspond to the project ID _(= 442, below)_:

image


NGitLab/Models/CommitStatus.cs
```diff
@@ -1,9 +1,11 @@
-﻿using System.Text.Json.Serialization;
+﻿using System;
+using System.Text.Json.Serialization;
 
 namespace NGitLab.Models
 {
     public class CommitStatus
     {
+        [Obsolete("This is NOT the 'Project Id' (could be some kind of 'Commit Status Id')")]
```
Toa741: As it's not a ProjectId at all, could we:
- JsonIgnore ProjectId + obsolete
- Add the 'Id' property with corresponding JsonPropertyName + comment to specify that it's the Id of the commit status if needed

NGitLab/Models/CommitCreate.cs
```diff
@@ -1,13 +1,13 @@
...
 using System.ComponentModel.DataAnnotations;
 using System.Text.Json.Serialization;
 
 namespace NGitLab.Models
 {
     public class CommitCreate
     {
-        [Required]
-        [JsonPropertyName("id")]
+        [Obsolete("'Project Id' is ignored in the POSTed JSON; it is actually specified through the endpoint URL.")]
```
Toa741: Can you add JsonIgnore on them too? To avoid sending 'projectId' or 'ProjectId' property

NGitLab/Models/EpicEdit.cs
```diff
@@ -1,12 +1,12 @@
...
+using System.ComponentModel.DataAnnotations;
 using System.Text.Json.Serialization;
 
 namespace NGitLab.Models
 {
     public class EpicEdit
     {
-        [Required]
-        [JsonPropertyName("id")]
+        [Obsolete("'Project Id' is ignored in the PUT JSON; it is actually specified through the endpoint URL.")]
```
Toa741:
```suggestion
        [Obsolete("'Id' is ignored in the PUT JSON; it is actually specified through the endpoint URL.")]
```

NGitLab/Models/IssueEdit.cs
```diff
@@ -8,7 +8,6 @@ namespace NGitLab.Models
     public class IssueEdit
     {
         [Required]
-        [JsonPropertyName("id")]
         public int Id;
```
Toa741: It seems that some of these 'Id' properties are in fact 'ProjectId's.

Should we also mark them obsolete and add a proper ProjectId property to make it clearer

NGitLab/Models/LabelEdit.cs
```diff
@@ -11,14 +12,19 @@ public LabelEdit()
 
         public LabelEdit(int projectId, Label label)
```
Toa741: It would probably be nicer if we had a proper distinction whether it's projectId or groupId.

Could we have like:
- ProjectLabelEdit that takes a projectId
- GroupLabelEdit that takes a groupId
- Both inherits LabelEdit
- All methods that takes a LabelEdit will then been enforced with the correct type

Toa741: After discussion:
- Avoid making a breaking change
- But create new class with new methods along
- Do not put things as obsolete, but use EditorBrowsableAttribute for "soft" obsolete

louis-z: Did most of what was suggested...

PR #409, opened 2023-03-15
Adding Issue.Clone API method

apoufortin:


NGitLab/Impl/IssueClient.cs
```diff
@@ -153,6 +154,12 @@ public Task<TimeStats> TimeStatsAsync(int projectId, int issueIid, CancellationT
             return _api.Get().ToAsync<TimeStats>(string.Format(CultureInfo.InvariantCulture, TimeStatsUrl, projectId, issueIid), cancellationToken);
         }
 
+        public Task<Issue> CloneAsync(int projectId, int issueIid, int toProjectId, bool withNotes = false, CancellationToken cancellationToken = default)
```
louis-z: You should create an `IssueClone` class and make `CloneAsync` take it in as parameter.

PR #426, opened 2023-04-18
Add cherry-pick support for commit

PMExtra:
https://docs.gitlab.com/ee/api/commits.html#cherry-pick-a-commit

### DIFF
--- a/NGitLab/Impl/CommitClient.cs
+++ b/NGitLab/Impl/CommitClient.cs
@@ -23,6 +23,11 @@ namespace NGitLab.Impl
             return _api.Get().To<Commit>(_repoPath + $"/commits/{@ref}");
         }
 
+        public Commit CherryPick(CommitCherryPick cherryPick)
+        {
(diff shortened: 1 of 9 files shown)


NGitLab/Models/CommitCherryPick.cs
```diff
@@ -0,0 +1,19 @@
+using System.ComponentModel.DataAnnotations;
+using System.Text.Json.Serialization;
+
+namespace NGitLab.Models
+{
+    public class CommitCherryPick
+    {
+        [Required]
+        [JsonIgnore]
+        public string Sha;
```
louis-z: Use the `Sha1` type instead.

NGitLab/Models/CommitCherryPick.cs
louis-z, line 1: All class members here should be properties, not fields.

NGitLab/Impl/CommitClient.cs
louis-z, line 1: Thanks @PMExtra for your contribution. it's appreciated.

Can you add at least 1 test that validates cherry-picking?

louis-z (changes requested):

louis-z (approved):

PR #433, opened 2023-04-28
Support searching branches

PMExtra:
https://docs.gitlab.com/15.11/ee/api/branches.html#list-repository-branches

```
GET /projects/:id/repository/branches
```

Attribute | Type | Required | Description
-- | -- | -- | --
search | string | no | Return list of branches containing the search string.<br/> You can use `^term` and `term$` to find branches that<br/> begin and end with term respectively.


NGitLab.Mock.Tests/BranchMockTests.cs
```diff
@@ -0,0 +1,39 @@
...
+                    .WithCommit("Initial commit")
+                    .WithCommit("Commit for branch_1", sourceBranch: "branch_1"))
+                .BuildServer();
+
+            var client = server.CreateClient("user1");
+            var branchClient = client.GetRepository(1).Branches;
+
+            var branches = branchClient.Search("main").ToList();
+            Assert.IsTrue(branches.Any(branch => string.Equals(branch.Name, "main", System.StringComparison.Ordinal)));
+            Assert.IsFalse(branches.Any(branch => string.Equals(branch.Name, "branch_1", System.StringComparison.Ordinal)));
```
Toa741:
```suggestion
            var branches = branchClient.Search("main");
            var expectedBranch = branches.Single();
            Assert.AreEqual("main", expectedBranch.Name);
```

As the scenarios should only return 1 branch, calling single() makes the assert part a bit more readable (need only one assert that avoid possible confusion with the IsTrue/IsFalse sequence).

The same can likely be applied in most cases in [BranchClientTests.cs](https://github.com/ubisoft/NGitLab/pull/433/files#diff-7bdcc99ac007d231c7852e392552593d988c95a3af8a8593764d37299fb47720)

PR #444, opened 2023-05-15
Add ability to do a code compare.

shivan:
I found that your library is maintained. But it misses the code compare feature. So I added it.

### DIFF
--- a/NGitLab.Tests/CompareTests.cs
+++ b/NGitLab.Tests/CompareTests.cs
@@ -63,7 +63,7 @@ namespace NGitLab.Tests
 
             Assert.Catch< GitLabException>(() =>
             {
-                var compareResults = context.Client.GetRepository(project.Id).Compare(new CompareQuery(project.DefaultBranch, "testblub"));
+                context.Client.GetRepository(project.Id).Compare(new CompareQuery(project.DefaultBranch, "testblub"));
             }, "404 Ref Not Found", null);
         }
     }
--- a/NGitLab.Tests/CompareTests.cs
+++ b/NGitLab.Tests/CompareTests.cs
@@ -16,7 +16,7 @@ namespace NGitLab.Tests
             var compareResults = context.Client.GetRepository(project.Id).Compare(new CompareQuery(project.DefaultBranch, project.DefaultBranch));
 
             Assert.IsNotNull(compareResults);
-            Assert.IsTrue(compareResults.Commits.Length == 0);
+            Assert.AreEqual(0, compareResults.Commits.Length);
         }
 
         [Test]
--- a/NGitLab.Tests/CompareTests.cs
+++ b/NGitLab.Tests/CompareTests.cs
@@ -1,0 +1,63 @@
+﻿using System.Threading.Tasks;
+using NGitLab.Models;
+using NGitLab.Tests.Docker;
+using NUnit.Framework;
+
+namespace NGitLab.Tests
+{
+    public class CompareTests
+    {
+        [Test]
+        [NGitLabRetry]
+        public async Task Test_compare_equal()
+        {
+            using var context = await GitLabTestContext.CreateAsync();
+            var project = context.CreateProject(initializeWithCommits: true);
+            var compareResults = context.Client.GetRepository(project.Id).Compare(new CompareQuery(project.DefaultBranch, project.DefaultBranch));
+
+            Assert.IsNotNull(compareResults);
+            Assert.IsTrue(compareResults.Commits.Length == 0);
+        }
+
+        [Test]
+        [NGitLabRetry]
+        public async Task Test_compare()
+        {
+            using var context = await GitLabTestContext.CreateAsync();
+            var project = context.CreateProject(initializeWithCommits: true);
+
+            context.Client.GetRepository(project.Id).Files.Create(new FileUpsert
+            {
+                Branch = "devtest",
+                CommitMessage = "file to be compared",
+                Path = "compare.txt",
+                RawContent = "compare me",
+            });
+
+            context.Client.GetRepository(project.Id).Files.Create(new FileUpsert
+            {
+                Branch = "devtest",
+                CommitMessage = "file to be compared, too",
+                Path = "compare.txt",
+                RawContent = "compare me now",
+            });
+
+            var compareResults = context.Client.GetRepository(project.Id).Compare(new CompareQuery(project.DefaultBranch, "devtest"));
+
+            Assert.IsNotNull(compareResults);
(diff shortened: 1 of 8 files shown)


NGitLab/IRepositoryClient.cs
```diff
@@ -48,5 +48,9 @@ public interface IRepositoryClient
         IBranchClient Branches { get; }
 
         IProjectHooksClient ProjectHooks { get; }
+
+        CompareInfo Compare(Sha1 from, Sha1 to);
+
+        CompareInfo Compare(string source, string target);
```
louis-z: What if we want to compare a commit against a branch or tag? We should probably always handle `from` and `to` as strings (i.e. provide a single `Compare` method) and let GitLab figure out what they represent.

Furthermore, I think that the single `Compare` method should take a "query" object instead, to reduce risks of breaking changes if we ever decide to support more query parameters:

```cs
public class CompareQuery
{
    /// <summary>
    /// Starting commit SHA, branch name or tag
    /// </summary>
    public string From { get; set; }
	
    /// <summary>
    /// Ending commit SHA, branch name or tag
    /// </summary>
    public string To { get; set; }
}
```

NGitLab/Models/CompareInfo.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System.Text.Json.Serialization;
+
+namespace NGitLab.Models
+{
+    public class CompareInfo
```
louis-z: I feel that something like `CompareResults` would be a better name.

louis-z (changes requested):

louis-z:
Thanks @shivan for your contribution. Could you also add a few tests, to validate what happens in different cases:
- `from` and `to` refer to valid and different commits
- `from` and `to` refer to the same commit
- `from` or `to` are invalid

shivan:
Hi @louis-z, I implemented your suggested changes. 
I also tried to implement some tests, but these did not run successfully. Seems like it needs somehow a docker container to run the tests.

louis-z:
> Hi @louis-z, I implemented your suggested changes. I also tried to implement some tests, but these did not run successfully. Seems like it needs somehow a docker container to run the tests.

Yes, you indeed need to have the Docker engine installed, as mentioned in our [README](https://github.com/ubisoft/NGitLab#running-unit-tests-locally).

NGitLab.Tests/CompareTests.cs
```diff
@@ -0,0 +1,63 @@
...
+        [Test]
+        [NGitLabRetry]
+        public async Task Test_compare()
+        {
+            using var context = await GitLabTestContext.CreateAsync();
+            var project = context.CreateProject(initializeWithCommits: true);
+
+            context.Client.GetRepository(project.Id).Files.Create(new FileUpsert
+            {
+                Branch = "devtest",
```
louis-z, line 37: Tests are failing because you're trying to commit to a missing branch. You need to create the branch beforehand.

NGitLab.Tests/CompareTests.cs
```diff
@@ -0,0 +1,63 @@
...
+        [Test]
+        [NGitLabRetry]
+        public async Task Test_compare_equal()
+        {
+            using var context = await GitLabTestContext.CreateAsync();
+            var project = context.CreateProject(initializeWithCommits: true);
+            var compareResults = context.Client.GetRepository(project.Id).Compare(new CompareQuery(project.DefaultBranch, project.DefaultBranch));
+
+            Assert.IsNotNull(compareResults);
+            Assert.IsTrue(compareResults.Commits.Length == 0);
```
louis-z:
```suggestion
            Assert.AreEqual(0, compareResults.Commits.Length);
```

NGitLab.Tests/CompareTests.cs
```diff
@@ -0,0 +1,63 @@
...
+                Branch = "devtest",
+                CommitMessage = "file to be compared, too",
+                Path = "compare.txt",
+                RawContent = "compare me now",
+            });
+
+            var compareResults = context.Client.GetRepository(project.Id).Compare(new CompareQuery(project.DefaultBranch, "devtest"));
+
+            Assert.IsNotNull(compareResults);
+            Assert.IsTrue(compareResults.Commits.Length == 2);
```
louis-z:
```suggestion
            Assert.AreEqual(2, compareResults.Commits.Length);
```

NGitLab.Tests/CompareTests.cs
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        [Test]
+        [NGitLabRetry]
+        public async Task Test_compare_invalid()
+        {
+            using var context = await GitLabTestContext.CreateAsync();
+            var project = context.CreateProject(initializeWithCommits: true);
+            var compareResults = context.Client.GetRepository(project.Id).Compare(new CompareQuery(project.DefaultBranch, "testblub"));
```
louis-z: This call throws, with a 404. The test should validate that.

NGitLab/Models/CompareQuery.cs
```diff
@@ -0,0 +1,24 @@
+﻿namespace NGitLab.Models
+{
+    /// <summary>
+    /// Query details for comparision of branches/tags/commit hashes
```
louis-z: Fix `comparision` -> `comparison` typo everywhere.

NGitLab/Impl/RepositoryClient.cs
```diff
@@ -96,6 +96,11 @@ public IEnumerable<Commit> GetCommits(GetCommitsRequest request)
             return allCommits.Take(request.MaxResults);
         }
 
+        public CompareResults Compare(CompareQuery query)
+        {
+            return _api.Get().To<CompareResults>(_repoPath + $@"/compare?from={query.Target}&to={query.Source}");
```
louis-z:
```suggestion
            return _api.Get().To<CompareResults>(_repoPath + $@"/compare?from={query.Source}&to={query.Target}");
```

louis-z (changes requested):

NGitLab/Models/CompareQuery.cs
shivan: done

NGitLab.Tests/CompareTests.cs
shivan: fixed

NGitLab.Tests/CompareTests.cs
shivan: validated now with exception

shivan:
@louis-z can you have a look again, I think I fixed everything (even got the tests running)

PR #451, opened 2023-05-26
Add project's permissions

jairbubbles:
See https://docs.gitlab.com/ee/api/projects.html

```
    "permissions": {
      "project_access": {
        "access_level": 10,
        "notification_level": 3
      },
      "group_access": {
        "access_level": 50,
        "notification_level": 3
      }
    },
```

### DIFF
--- a/NGitLab/Models/Permissions.cs
+++ b/NGitLab/Models/Permissions.cs
@@ -1,0 +1,13 @@
+using System.Text.Json.Serialization;
+
+namespace NGitLab.Models
+{
+    public class Permissions
+    {
+        [JsonPropertyName("project_access")]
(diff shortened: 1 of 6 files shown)


NGitLab/Models/Permissions.cs
```diff
@@ -0,0 +1,13 @@
+using System.Text.Json.Serialization;
+
+namespace NGitLab.Models
+{
+    public class Permissions
```
Toa741: Non generic objects related to Project are prefix with it (ProjectStatistics).

Can we keep the same principle for permission(s)? As it might be generic enough to collide with something else at some point

jairbubbles: Sure!

PR #456, opened 2023-06-02
Add Bot property in User mock

Toa741:


NGitLab.Mock/User.cs
```diff
@@ -34,6 +34,8 @@ public User(string userName)
 
         public string WebUrl => Server.MakeUrl(UserName);
 
+        public bool Bot { get; set; }
```
meziantou: I don't think this should be settable. Instead, I would prefer a method such as

````c#
IGitLabClient Project.CreateProjectAccessToken();
````

This way, you can set all properties as indicated in the documentation https://docs.gitlab.com/ee/user/project/settings/project_access_tokens.html#bot-users-for-projects

Toa741: That'll be neater, will add it

PR #460, opened 2023-06-05
Add support for Lint endpoints

mchanyeechoy:
Add endpoints from https://docs.gitlab.com/ee/api/lint.html

### DIFF
--- a/NGitLab/Impl/LintClient.cs
+++ b/NGitLab/Impl/LintClient.cs
@@ -17,11 +17,12 @@ namespace NGitLab.Impl
         public Task<LintCI> ValidateCIYamlContentAsync(string projectId, string yamlContent, LintCIOptions options, CancellationToken cancellationToken = default)
         {
             var url = BuildLintCIUrl(projectId, options);
-
-            return _api.Post().With(new
+            var data = new
             {
                 content = yamlContent,
-            }).ToAsync<LintCI>(url, cancellationToken);
+            };
+
+            return _api.Post().With(data).ToAsync<LintCI>(url, cancellationToken);
         }
 
         public Task<LintCI> ValidateProjectCIConfigurationAsync(string projectId, LintCIOptions options, CancellationToken cancellationToken = default)
--- a/NGitLab/Impl/LintClient.cs
+++ b/NGitLab/Impl/LintClient.cs
@@ -1,0 +1,43 @@
+﻿using System.Net;
+using System.Threading;
+using System.Threading.Tasks;
+using NGitLab.Models;
+
+namespace NGitLab.Impl
+{
+    internal class LintClient : ILintClient
+    {
+        private readonly API _api;
+
+        public LintClient(API api)
+        {
+            _api = api;
+        }
+
+        public Task<LintCI> LintGitLabCIYamlAsync(string projectId, string yamlContent, LintCIOptions options, CancellationToken cancellationToken = default)
+        {
+            var url = BuildLintCIUrl(projectId, options);
+            url = Utils.AddParameter(url, "content", yamlContent);
+
+            return _api.Post().ToAsync<LintCI>(url, cancellationToken);
+        }
+
+        public Task<LintCI> LintProjectGitLabCIYamlAsync(string projectId, LintCIOptions options, CancellationToken cancellationToken = default)
+        {
+            var url = BuildLintCIUrl(projectId, options);
+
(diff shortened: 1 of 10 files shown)


NGitLab/Impl/LintClient.cs
```diff
@@ -0,0 +1,43 @@
...
+    internal class LintClient : ILintClient
+    {
+        private readonly API _api;
+
+        public LintClient(API api)
+        {
+            _api = api;
+        }
+
+        public Task<LintCI> LintGitLabCIYamlAsync(string projectId, string yamlContent, LintCIOptions options, CancellationToken cancellationToken = default)
```
louis-z: I think the methods can have clearer names.

In [the doc](https://docs.gitlab.com/ee/api/lint.html#validate-a-ci-yaml-configuration-with-a-namespace), this is described as `Validate a CI YAML configuration with a namespace`, but to make it even clearer that we're passing in YAML content:

```suggestion
        public Task<LintCI> ValidateCIYamlContentAsync(string projectId, string yamlContent, LintCIOptions options, CancellationToken cancellationToken = default)
```

NGitLab/Impl/LintClient.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            var url = BuildLintCIUrl(projectId, options);
+            url = Utils.AddParameter(url, "content", yamlContent);
+
+            return _api.Post().ToAsync<LintCI>(url, cancellationToken);
+        }
+
+        public Task<LintCI> LintProjectGitLabCIYamlAsync(string projectId, LintCIOptions options, CancellationToken cancellationToken = default)
```
louis-z: The [doc](https://docs.gitlab.com/ee/api/lint.html#validate-a-projects-ci-configuration) calls this `Validate a project’s CI configuration`

```suggestion
        public Task<LintCI> ValidateProjectCIConfigurationlAsync(string projectId, LintCIOptions options, CancellationToken cancellationToken = default)
```

NGitLab/Impl/LintClient.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        public LintClient(API api)
+        {
+            _api = api;
+        }
+
+        public Task<LintCI> LintGitLabCIYamlAsync(string projectId, string yamlContent, LintCIOptions options, CancellationToken cancellationToken = default)
+        {
+            var url = BuildLintCIUrl(projectId, options);
+            url = Utils.AddParameter(url, "content", yamlContent);
```
louis-z: According to the request example in the doc, `yamlContent` is supposed to be passed in as POST data, not as query parameter:

> curl --header "Content-Type: application/json" "https://gitlab.example.com/api/v4/projects/:id/ci/lint" --data '{"content": "{ \"image\": \"ruby:2.6\", \"services\": [\"postgres\"], \"before_script\": [\"bundle install\", \"bundle exec rake db:create\"], \"variables\": {\"DB_NAME\": \"postgres\"}, \"types\": [\"test\", \"deploy\", \"notify\"], \"rspec\": { \"script\": \"rake spec\", \"tags\": [\"ruby\", \"postgres\"], \"only\": [\"branches\"]}}"}'

It's strange that tests are passing... 🤔

NGitLab/Models/LintCI.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace NGitLab.Models
+{
+    public class LintCI
+    {
+        public const string Url = "/ci/lint";
+
+        [JsonPropertyName("valid")]
+        public bool Valid { get; set; }
+
+        [JsonPropertyName("id")]
```
louis-z:
```suggestion
        [JsonPropertyName("merged_yaml")]
```

NGitLab/Models/LintCIOptions.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace NGitLab.Models
+{
+    public class LintCIOptions
+    {
+        /// <summary>
+        /// Run pipeline creation simulation, or only do static check.
+        /// </summary>
+        public bool DryRun { get; set; }
```
louis-z: Since it's not a required param, can you make it nullable?

louis-z (changes requested):

NGitLab/Impl/LintClient.cs
mchanyeechoy: Seems like they both work, but I will change it to be inline with the docs 🙂

louis-z (approved):

PR #461, opened 2023-06-05
Add GitLabTestContext.ReportTestAsInconclusiveIfVersionOutOfRange

louis-z:
Centralize the GitLab version checking logic that reports tests as inconclusive. It was being previously repeated across tests.

### DIFF
--- a/NGitLab.Tests/Docker/GitLabTestContext.cs
+++ b/NGitLab.Tests/Docker/GitLabTestContext.cs
@@ -11,6 +11,7 @@ using System.Runtime.InteropServices;
 using System.Security.Cryptography;
 using System.Threading;
 using System.Threading.Tasks;
+using Meziantou.Framework.Versioning;
 using NGitLab.Models;
 using NUnit.Framework;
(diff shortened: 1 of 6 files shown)


NGitLab.Tests/Docker/GitLabTestContext.cs
```diff
@@ -254,6 +255,19 @@ public int GetRandomNumber()
             return RandomNumberGenerator.GetInt32(0, int.MaxValue);
         }
 
+        public void ReportTestAsInconclusiveIfVersionOutOfRange(SemanticVersion minVersion = null, SemanticVersion maxVersion = null)
```
meziantou: I think it would be better to use a package such as NuGet.Versioning and use a VersionRange.
- It is clearer if min/max are inclusive or exclusive
- It would simplify `13.10.99` as you can set something like `[,13.11)`

```suggestion
        public void ReportTestAsInconclusiveIfVersionOutOfRange(VersionRange range)
```

PR #463, opened 2023-06-06
Run tests on GitLab 15.4.6 (default) & 15.11.6

louis-z:
Some of the tests that pass on GitLab 15.4.6 fail on GitLab 15.11.6, most of them related to MR creation. The pattern seems to be
- Create a merge request
- Immediately thereafter, try to retrieve from GitLab the said merge request via a specific commit
- End up with no results 😞 

Add calls to `GitLabTestContext.RetryUntilAsync` to wait for GitLab to fully assess a recently created MR's state.

### DIFF
--- a/NGitLab.Tests/CommitsTests.cs
+++ b/NGitLab.Tests/CommitsTests.cs
@@ -70,8 +70,10 @@ namespace NGitLab.Tests
                 SourceBranch = "test-mr",
                 TargetBranch = project.DefaultBranch,
                 Title = mergeRequestTitle,
-                AwaitAssessment = true,
             });
+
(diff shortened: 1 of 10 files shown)


NGitLab.Tests/CommitsTests.cs
```diff
@@ -72,6 +72,9 @@ public async Task Test_can_get_merge_request_associated_to_commit()
                 Title = mergeRequestTitle,
             });
 
+            // Sleep so GitLab has time to finish assessing the just created MR (otherwise the following calls will return nothing)
+            await Task.Delay(5000);
```
meziantou: I think you could use `context.RetryUntilAsync(...)`

PR #464, opened 2023-06-09
Remove redundant `projectId` from ContributorClient

PMExtra:
https://docs.gitlab.com/ee/api/repositories.html#contributors

The `id` argument has already been included in the `repoPath`.

No need to pass it through query parameters.


PMExtra:
The existing unit tests can validate this change without any modifications:
- `ContributorsTests.Test_can_get_contributors`
- `ContributorsTests.Test_can_get_MultipleContributors`

PR #485, opened 2023-07-10
Use PackageLicenseExpression

### LINKED ISSUE #480: Missing License Information In Nuget metadata
Hi,
Could the license information be added to the nuget information?
it is good to add it so it will be shown in Visual Studio nuget management.

for example,I use the following ```<PackageLicenseExpression>MIT</PackageLicenseExpression>``` in my nugets:

	<PropertyGroup>
		<PublishRepositoryUrl>true</PublishRepositoryUrl>
		<IncludeSymbols>true</IncludeSymbols>
		<SymbolPackageFormat>snupkg</SymbolPackageFormat>
		<ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
		<PackageLicenseExpression>MIT</PackageLicenseExpression>
		<RepositoryType>git</RepositoryType>
		<Nullable>enable</Nullable>
		<LangVersion>latest</LangVersion>
	</PropertyGroup>

Also some tools can create auto generated report like (dotnet-project-licenses)  but right now those create missing entry:
image

meziantou:
Fix https://github.com/ubisoft/NGitLab/issues/480

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -16,7 +16,7 @@
     <RestoreUseStaticGraphEvaluation>true</RestoreUseStaticGraphEvaluation>
 
     <!-- Package info -->
-    <PackageLicenseFile>LICENSE</PackageLicenseFile>
+    <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <PackageReadmeFile>README.md</PackageReadmeFile>
   </PropertyGroup>
 


Directory.Build.props
```diff
@@ -16,7 +16,7 @@
     <RestoreUseStaticGraphEvaluation>true</RestoreUseStaticGraphEvaluation>
 
     <!-- Package info -->
-    <PackageLicenseFile>LICENSE</PackageLicenseFile>
```
louis-z, line 19: With this change do we still need the `LICENSE` file?

meziantou: - We need it in the git repository
- I think we need it in the package as we forked from an MIT project, we need to copy the license file, but I'm not a licensing expert...

    > The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

PR #488, opened 2023-07-15
Add LinkedTo method to IIssueClient Interface

### LINKED ISSUE #482: [Question] get related Issues for specific Issue.
Hi,
Is there a way to get all related issues for a specific issues?
I normaly uses primary issues with "sub issues" linked to that one so this option to get it will be usefull to me for reporting.

LiorBanai:
## API Changes:

Adding the following method:
```cs
        /// <summary>
        /// Get all Issues that are linked to a particular issue of particular project.
        /// </summary>
        /// <param name="projectId">The project id.</param>
        /// <param name="issueId">The id of the issue in the project's scope.</param>
        /// <returns>The list of Issues linked to this issue.</returns>
        IEnumerable<Issue> LinkedTo(int projectId,int issueId);
```

## Open Issues:
- [x] Public API analyzer: needs action to resolve finding.
- [x] Mock class returns empty collection:
  ```cs
     public IEnumerable<Models.Issue> LinkedTo(int projectId, int issueId)
     {
         {
             return Enumerable.Empty<Models.Issue>();
         }
     }
  ```

## Tests:

- Tested locally on personal repository. Linked issues are fetched correctly 
- recursive test (linked issues of fetched linked issues) probably should be blocked due to ciruclar reference?

## Related Issues:

Fix #482

### DIFF
--- a/NGitLab.Mock/Clients/IssueClient.cs
+++ b/NGitLab.Mock/Clients/IssueClient.cs
@@ -302,6 +302,14 @@ namespace NGitLab.Mock.Clients
             return GetById(issueId);
         }
 
+        public IEnumerable<Models.Issue> LinkedTo(int projectId, int issueId)
+        {
+            using (Context.BeginOperationScope())
+            {
+                return Enumerable.Empty<Models.Issue>();
+            }
+        }
+
         public Models.Issue GetById(int issueId)
         {
             using (Context.BeginOperationScope())
--- a/NGitLab.Mock/Clients/IssueClient.cs
+++ b/NGitLab.Mock/Clients/IssueClient.cs
@@ -193,9 +193,7 @@ namespace NGitLab.Mock.Clients
         {
             throw new NotImplementedException();
         }
-
-
-
+        
         public GitLabCollectionResponse<Models.MergeRequest> RelatedToAsync(int projectId, int issueIid)
         {
             throw new NotImplementedException();
--- a/NGitLab.Mock/Clients/IssueClient.cs
+++ b/NGitLab.Mock/Clients/IssueClient.cs
@@ -302,12 +302,9 @@ namespace NGitLab.Mock.Clients
             return GetById(issueId);
         }
 
-        public IEnumerable<Models.Issue> LinkedTo(int projectId, int issueId)
+        public GitLabCollectionResponse<Models.Issue> LinkedTo(int projectId, int issueId)
         {
-            using (Context.BeginOperationScope())
-            {
-                return Enumerable.Empty<Models.Issue>();
-            }
+            throw new NotImplementedException();
         }
 
         public Models.Issue GetById(int issueId)
--- a/NGitLab.Mock/Clients/IssueClient.cs
+++ b/NGitLab.Mock/Clients/IssueClient.cs
@@ -193,7 +193,6 @@ namespace NGitLab.Mock.Clients
         {
             throw new NotImplementedException();
         }
-        
         public GitLabCollectionResponse<Models.MergeRequest> RelatedToAsync(int projectId, int issueIid)
         {
             throw new NotImplementedException();
--- a/NGitLab.Mock/Clients/IssueClient.cs
+++ b/NGitLab.Mock/Clients/IssueClient.cs
@@ -302,7 +302,7 @@ namespace NGitLab.Mock.Clients
             return GetById(issueId);
         }
 
-        public GitLabCollectionResponse<Models.Issue> LinkedTo(int projectId, int issueId)
+        public GitLabCollectionResponse<Models.Issue> LinkedToAsync(int projectId, int issueId)
         {
             throw new NotImplementedException();
         }
--- a/NGitLab.Mock/Clients/IssueClient.cs
+++ b/NGitLab.Mock/Clients/IssueClient.cs
@@ -194,6 +194,8 @@ namespace NGitLab.Mock.Clients
             throw new NotImplementedException();
         }
 
+
+
         public GitLabCollectionResponse<Models.MergeRequest> RelatedToAsync(int projectId, int issueIid)
         {
             throw new NotImplementedException();
@@ -303,6 +305,11 @@ namespace NGitLab.Mock.Clients
         }
 
         public GitLabCollectionResponse<Models.Issue> LinkedTo(int projectId, int issueId)
+        {
+            throw new NotImplementedException();
+        }
+
+        public bool CreateLinkBetweenIssues(int sourceProjectId, int sourceIssueId, int targetProjectId, int targetIssueId)
         {
             throw new NotImplementedException();
         }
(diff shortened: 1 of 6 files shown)


NGitLab/IIssueClient.cs
```diff
@@ -134,6 +134,14 @@ public interface IIssueClient
         /// <returns>The list of MR that are related this issue.</returns>
         IEnumerable<MergeRequest> RelatedTo(int projectId, int issueIid);
 
+        /// <summary>
+        /// Get all Issues that are linked to a particular issue of particular project.
+        /// </summary>
+        /// <param name="projectId">The project id.</param>
+        /// <param name="issueId">The id of the issue in the project's scope.</param>
+        /// <returns>The list of Issues linked to this issue.</returns>
+        IEnumerable<Issue> LinkedTo(int projectId,int issueId);
```
meziantou: You should use the newer type `GitLabCollectionResponse` which allow sync and async usage.

```suggestion
        GitLabCollectionResponse<Issue> LinkedTo(int projectId,int issueId);
```

NGitLab.Mock/Clients/IssueClient.cs
```diff
@@ -302,6 +302,14 @@ public async Task<Models.Issue> GetByIdAsync(int issueId, CancellationToken canc
             return GetById(issueId);
         }
 
+        public IEnumerable<Models.Issue> LinkedTo(int projectId, int issueId)
+        {
+            using (Context.BeginOperationScope())
+            {
+                return Enumerable.Empty<Models.Issue>();
```
meziantou: If you can't implement it, you should make it clear the method is not implemented:

```suggestion
                throw new NotImplementedException();
```

meziantou (changes requested):
Could you also add a test to validate the method is correctly implemented in https://github.com/ubisoft/NGitLab/blob/main/NGitLab.Tests/IssueTests.cs?

This is quite important as the behavior of GitLab can change when you upgrade the server. So, the tests allow us to detect breaking changes.

NGitLab.Mock/Clients/IssueClient.cs
LiorBanai: Sure, I'll take care of all those raised issues

NGitLab/IIssueClient.cs
LiorBanai: BTW, the new ```LinkedTo``` uses the same pattern as ```RelatedTo```:
image

meziantou: The GitLabCollectionResponse is recent, so you can see lots of methods that still use the old pattern.

LiorBanai:
> Could you also add a test to validate the method is correctly implemented in https://github.com/ubisoft/NGitLab/blob/main/NGitLab.Tests/IssueTests.cs?
> 
> This is quite important as the behavior of GitLab can change when you upgrade the server. So, the tests allow us to detect breaking changes.

the ```Issue``` class has no Linked property so not sure what the test logic should be. basically the link https://docs.gitlab.com/ee/api/issue_links.html#list-issue-relations
just fetch list of issues that related to a specific issue. There is no array property in the ```Issue``` model.

meziantou:
Maybe I'm wrong, but does the following flow works:
- Create 1 issue
- Create another issue with the description `blocks #1`
- Get all links for the first issues => The collection must contain 1 item

meziantou:
The build fails. You need to fix the public API file. You can use [Quick Actions](https://learn.microsoft.com/en-us/visualstudio/ide/quick-actions?view=vs-2022&WT.mc_id=DT-MVP-5003978) in VS to update the file and fix the warning.

LiorBanai:
Added the unit test and updarted the public API file

meziantou (approved):

LiorBanai:
Hi,
I hve no idea why the unit test fails and don't reallyhave a way to test it locally.

LiorBanai:
Ok , it seems linking by description does not work even manually (I created manually 2 issues and try linking by comments/description).
it does not work.

But we can Add post command to link issues:
https://docs.gitlab.com/ee/api/issue_links.html#create-an-issue-link

LiorBanai:
But adding ability to Create Linked Issues should be different PR...

LiorBanai:
the trx show 200 result for get linked:
```
----------------------------------------------------------------------------------------------------
GitLab request: GET http://localhost:48624/api/v4/projects/59/issues/1/links
Accept-Encoding: gzip
Private-Token: ****** 
User-Agent: NGitLab.Tests/1.0.0
----------
200 OK
Server: nginx
Date: Tue, 18 Jul 2023 21:17:09 GMT
Connection: keep-alive
Cache-Control: must-revalidate, max-age=0, private
ETag: W/"4f53cda18c2baa0c0354bb5f9a3ecbe5"
Vary: Origin
X-Content-Type-Options: nosniff
X-Frame-Options: SAMEORIGIN
X-Request-ID: 01H5NDTD001NG2C2AM5W4B9THZ
X-Runtime: 0.074998
Strict-Transport-Security: max-age=63072000
Referrer-Policy: strict-origin-when-cross-origin
Content-Type: application/json
Content-Length: 2

[]
```

LiorBanai:
Added second new API:
```cs
        /// <summary>
        /// Create links between Issues.
        /// </summary>
        /// <param name="sourceProjectId">The project id.</param>
        /// <param name="sourceIssueId">The id of the issue in the project's scope.</param>
        /// <param name="targetProjectId">The target project id.</param>
        /// <param name="targetIssueId">The target id of the issue to link to.</param>
        bool CreateLinkBetweenIssues(int sourceProjectId, int sourceIssueId, int targetProjectId, int targetIssueId);

```

with this api we can update the test to create and get linked issues.
Tested both API on my presonal Gitlab repository

meziantou:
Looks good! 

You need to fix the Stylecop error and rename the new method with the "Async" suffix to fix the test.

LiorBanai:
I'll fix those. Why aren't those set as error in the .editorconfig?

meziantou:
> Why aren't those set as error in the .editorconfig?

- For the analyzer: they are available locally, but only as warning. On CI, it uses `TreatWarningsAsErrors`, so it fails the build.
- For the `Async` suffix, I'm not sure it could be part of the naming convention (editorconfig) as it only concerns methods returning `GitLabCollectionResponse`. Also, it hard to apply rules in this old and non-consistent codebase. We don't want to introduce breaking changes, nor too many warnings.

LiorBanai:
Updated the public api again . Hopefuly last commit :)

NGitLab.Mock/Clients/IssueClient.cs
```diff
@@ -193,7 +193,7 @@ public IEnumerable<Models.MergeRequest> RelatedTo(int projectId, int issueId)
         {
             throw new NotImplementedException();
         }
-
+
```
meziantou: Maybe not the last commit ;)

```suggestion

```

meziantou:

PR #514, opened 2023-08-31
Implement "IncludeInheritedMembers" in Membership client mock

mlemieuxlafontaine-ubi:


NGitLab.Mock/Clients/MembersClient.cs
```diff
@@ -98,8 +98,15 @@ public IEnumerable<Membership> OfGroup(string groupId, bool includeInheritedMemb
             using (Context.BeginOperationScope())
             {
                 var group = GetGroup(groupId, GroupPermission.View);
-                var members = group.GetEffectivePermissions().Permissions;
-                return members.Select(member => member.ToMembershipClient());
+                if (includeInheritedMembers)
```
louis-z: I think you could have simply just passed `includeInheritedMembers` to `group.GetEffectivePermissions()` and have the latter handle inherited members.

mlemieuxlafontaine-ubi: Hmm, that's an interesting approach. I'll check to see if it looks better yes

mlemieuxlafontaine-ubi: It sure did look better, and I think allows us to more closely replicate GitLab's behavior

PR #515, opened 2023-09-04
Implements since and until parameters when retrieving commit list

steve85:
Implements the ability to use the _since_ and _until_ parameters when retrieving a list of commits.


steve85:
Closing this PR as I accidentally committed using incorrect author details. Will re-create a new one.

PR #516, opened 2023-09-04
Implements since and until parameters when retrieving commit list

steve85:
This PR adds two new properties to the [GitCommitsRequest](https://github.com/ubisoft/NGitLab/pull/516/files#diff-40287b0782fbac6570425266f20834bfb7732ec77a218b1721d92024455e565a) class, _Since_ and _Until_. The purpose of this is to allow callers to retrieve a list of commits within a specified date range.

GitLab API reference for listing commits: https://docs.gitlab.com/ee/api/commits.html#list-repository-commits

### DIFF
--- a/NGitLab/Impl/RepositoryClient.cs
+++ b/NGitLab/Impl/RepositoryClient.cs
@@ -1,7 +1,9 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Net;
 using NGitLab.Extensions;
 using NGitLab.Models;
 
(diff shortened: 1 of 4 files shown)


NGitLab/Impl/RepositoryClient.cs
```diff
@@ -85,6 +87,16 @@ public IEnumerable<Commit> GetCommits(GetCommitsRequest request)
                 lst.Add($"first_parent={Uri.EscapeDataString(request.FirstParent.ToString())}");
             }
 
+            if (request.Since.HasValue)
+            {
+                lst.Add($"since={WebUtility.UrlEncode(request.Since.Value.ToString("s", CultureInfo.InvariantCulture))}");
```
meziantou: NGitLab uses `Uri.EscapeDataString` everywhere to escape query string value. Can you please use it?

meziantou (changes requested):

NGitLab/Impl/RepositoryClient.cs
steve85: Sure thing, I will push a change for this shortly.

meziantou (approved):

PR #524, opened 2023-09-11
Ability to upload to Generic Packages Repository

gep13:


gep13:
@meziantou I have opened this as draft, based on the comments that you have made here:

https://github.com/ubisoft/NGitLab/discussions/511#discussioncomment-6917731

I would love to get some feedback on what I have done here, and what changes/suggestions you could make.

This is quite a naive implementation based on my own needs, and the information that I could find here:

https://docs.gitlab.com/ee/user/packages/generic_packages/index.html#do-not-allow-duplicate-generic-packages

Thanks!

NGitLab/Models/PackagePublish.cs
```diff
@@ -0,0 +1,30 @@
+using System.ComponentModel.DataAnnotations;
+using System.Text.Json.Serialization;
+
+namespace NGitLab.Models
+{
+    public class PackagePublish
+    {
+        [Required]
+        [JsonPropertyName("id")]
+        public int ProjectId;
```
meziantou: New files should use getter and setter `{ get; set; }`

NGitLab/IPackageClient.cs
```diff
@@ -0,0 +1,14 @@
...
+namespace NGitLab
+{
+    public interface IPackageClient
+    {
+        /// <summary>
+        /// Add a package file with the proposed information to the GitLab Generic Package Repository for the selected Project Id.
+        /// </summary>
+        /// <param name="packagePublish">The information about the package file to publish.</param>
+        /// <returns>The package if it was created.  Null if not.</returns>
+        Package Publish(PackagePublish packagePublish);
```
meziantou: New methods must be async. 

```suggestion
        Task<Package> PublishAsync(PackagePublish packagePublish, CancellationToken cancellationToken = default);
```

NGitLab/Impl/PackageClient.cs
```diff
@@ -0,0 +1,36 @@
...
+            if (!File.Exists(fullFilePath))
+            {
+                throw new InvalidOperationException("Can't find file!");
+            }
+
+            var formData = new FileContent(File.OpenRead(fullFilePath));
+            var packageFile = new FileInfo(fullFilePath);
+
+            return _api.Put().With(formData).To<Package>(string.Format(CultureInfo.InvariantCulture,
+                PublishPackageUrl, packagePublish.ProjectId, packagePublish.PackageName, packagePublish.PackageVersion,
```
meziantou: Values must be encoded with `Uri.EscapeDataString`

NGitLab/Impl/PackageClient.cs
```diff
@@ -0,0 +1,36 @@
...
+        private readonly API _api;
+
+        public PackageClient(API api)
+        {
+            _api = api;
+        }
+
+        public Package Publish(PackagePublish packagePublish)
+        {
+            var fullFilePath = Path.GetFullPath(packagePublish.FileName);
```
meziantou: I don't think this code should manipulate the file system. Instead `PackagePublish` should expose a `Stream`. If needed, you can add a helper method to set all properties from a path.

NGitLab/Models/PackageFile.cs
```diff
@@ -0,0 +1,10 @@
+using System.Text.Json.Serialization;
+
+namespace NGitLab.Models
+{
+    public class PackageFile
+    {
+        [JsonPropertyName("url")]
```
meziantou, line 7: New files should use getter and setter `{ get; set; }`

NGitLab/Models/Package.cs
```diff
@@ -0,0 +1,76 @@
...
+namespace NGitLab.Models
+{
+    public class Package
+    {
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        [JsonIgnore]
+        public int Id;
+
+        [JsonPropertyName("package_id")]
+        public int PackageId;
```
meziantou: New files should use getter and setter `{ get; set; }`

NGitLab/Models/Package.cs
```diff
@@ -0,0 +1,76 @@
...
+using System.Text.Json.Serialization;
+using NGitLab.Impl.Json;
+
+namespace NGitLab.Models
+{
+    public class Package
+    {
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        [JsonIgnore]
+        public int Id;
```
meziantou: I know some endpoints expose URL parameters using JsonIgnore properties, but it is for legacy reasons. It seems better to expose this value as a method parameter `public Package Publish(int projectId, PackagePublish packagePublish)`

NGitLab.Tests/PackageTests.cs
```diff
@@ -0,0 +1,33 @@
...
+                FileName = "../../../../README.md",
+                PackageName = "Packages",
+                PackageVersion = "1.0.0",
+                Status = "default",
+            };
+
+            var package1 = packagesClient.Publish(packagePublish);
+
+            // TODO: What assertions should be put here?
+            Assert.True(true);
```
meziantou: You should implement https://docs.gitlab.com/ee/api/packages.html#get-a-project-package, so you can fetch the package after publishing it and assert the package is available

meziantou (changes requested):

meziantou:
> I would love to get some feedback on what I have done here, and what changes/suggestions you could make.

That's a good start! I've added some comments, but the shape is ok

NGitLab.Mock/Clients/PackageClient.cs
```diff
@@ -0,0 +1,30 @@
...
+            : base(context)
+        {
+        }
+
+        public Task<Package> PublishAsync(int projectId, PackagePublish packagePublish, CancellationToken cancellationToken = default)
+        {
+            throw new System.NotImplementedException();
+        }
+
+        public IEnumerable<PackageSearchResult> Get(int projectId, PackageQuery packageQuery)
```
gep13, line 20: @meziantou I couldn't see any prior art for returning an `IEnumerable<T>` in any of the existing clients.  Happy to make a change here, just wanted to make sure that we are on the same page before making any changes.

gep13:
@meziantou I have taken a stab at making the changes that you suggested, and I have implemented a couple more methods on the new PackageClient for `Get` and `GetByIdAsync`.  Let me know if you have any further questions/suggestions.

Side Question... Should I create an issue for this feature addition, or assuming that we move forward with this, are you ok to merge the PR directly, without an associated issue?

NGitLab.Mock/Clients/PackageClient.cs
meziantou: Other implementations use `GitLabCollectionResponse<T>`. This supports sync and async enumeration!

NGitLab/Models/Package.cs
```diff
@@ -0,0 +1,74 @@
+using System;
+using System.Text.Json.Serialization;
+using NGitLab.Impl.Json;
+
+namespace NGitLab.Models
+{
+    public class Package
+    {
+        [JsonPropertyName("id")]
+        public int Id { get; set; }
```
meziantou, line 10: New classes should use `long` for ids

```suggestion
        public long Id { get; set; }
```

NGitLab/Models/Package.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        [JsonPropertyName("package_id")]
+        public int PackageId { get; set; }
```
meziantou, line 13:
```suggestion
        public long PackageId { get; set; }
```

NGitLab/Models/Package.cs
```diff
@@ -0,0 +1,74 @@
...
+        [JsonPropertyName("created_at")]
+        [JsonConverter(typeof(DateOnlyConverter))]
+        public DateTime CreatedAt { get; set; }
+
+        [JsonPropertyName("updated_at")]
+        [JsonConverter(typeof(DateOnlyConverter))]
+        public DateTime? UpdatedAt { get; set; }
+
+        [JsonPropertyName("size")]
+        public int Size { get; set; }
```
meziantou, line 24: File size is a long. I'm not sure GitLab would allow such big files, but you never know

```suggestion
        public long Size { get; set; }
```

NGitLab/Models/Package.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        [JsonPropertyName("file_store")]
+        public int FileStore { get; set; }
+
+        [JsonPropertyName("file_md5")]
+        public string FileMD5 { get; set; }
```
meziantou, line 30:
```suggestion
        public string FileMd5 { get; set; }
```

NGitLab/Models/Package.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        [JsonPropertyName("file_sha1")]
+        public string FileSHA1 { get; set; }
```
meziantou, line 33:
```suggestion
        public string FileSha1 { get; set; }
```

NGitLab/Models/Package.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        [JsonPropertyName("file_sha256")]
+        public string FileSHA256 { get; set; }
```
meziantou, line 36:
```suggestion
        public string FileSha256 { get; set; }
```

NGitLab/Models/PackageQuery.cs
```diff
@@ -0,0 +1,23 @@
+namespace NGitLab.Models
+{
+    public class PackageQuery
+    {
+        public int PackageId { get; set; }
```
meziantou, line 5:
```suggestion
        public long PackageId { get; set; }
```

NGitLab/Models/PackageSearchResult.cs
```diff
@@ -0,0 +1,35 @@
+using System;
+using System.Text.Json.Serialization;
+using NGitLab.Impl.Json;
+
+namespace NGitLab.Models
+{
+    public class PackageSearchResult
+    {
+        [JsonPropertyName("id")]
+        public int PackageId { get; set; }
```
meziantou, line 10:
```suggestion
        public long PackageId { get; set; }
```

NGitLab/Models/PackageStatus.cs
```diff
@@ -0,0 +1,11 @@
+namespace NGitLab.Models
+{
+    public enum PackageStatus
```
meziantou, line 3: Use C# name (e.g. `Default`) and the `EnumMember(Value = "")]` attribute to map the actual name

NGitLab/Models/PackageType.cs
```diff
@@ -0,0 +1,17 @@
+namespace NGitLab.Models
+{
+    public enum PackageType
+    {
+        all,
```
meziantou, line 5: Use C# name (e.g. `Default`) and the `EnumMember(Value = "")]` attribute to map the actual name

NGitLab/Models/PackageSort.cs
```diff
@@ -0,0 +1,8 @@
+namespace NGitLab.Models
+{
+    public enum PackageSort
+    {
+        asc,
```
meziantou, line 5: Use C# name (e.g. `Default`) and the `EnumMember(Value = "")]` attribute to map the actual name

NGitLab/Impl/PackageClient.cs
```diff
@@ -0,0 +1,72 @@
...
+        private const string GetPackageUrl = "/projects/{0}/packages/{1}";
+
+        private readonly API _api;
+
+        public PackageClient(API api)
+        {
+            _api = api;
+        }
+
+        public Task<Package> PublishAsync(int projectId, PackagePublish packagePublish, CancellationToken cancellationToken = default)
```
meziantou, line 23: Based on the url, I think the method name should be `PublishGenericPackageAsync`

```suggestion
        public Task<Package> PublishGenericPackageAsync(int projectId, PackagePublish packagePublish, CancellationToken cancellationToken = default)
```

NGitLab.Tests/PackageTests.cs
```diff
@@ -0,0 +1,40 @@
...
+            var project = context.CreateProject();
+            var packagesClient = context.Client.Packages;
+
+            var packagePublish = new PackagePublish
+            {
+                FileName = "README.md",
+                PackageName = "Packages",
+                PackageVersion = "1.0.0",
+                Status = "default",
+                PackageStream = File.OpenRead("../../../../README.md"),
```
meziantou, line 26: You'll need to dispose the stream. Also, don't rely on the file system, you can use a `MemoryStream` to create a dummy content.

meziantou:
> Side Question... Should I create an issue for this feature addition, or assuming that we move forward with this, are you ok to merge the PR directly, without an associated issue?

No need to create an issue

NGitLab.Mock/Clients/PackageClient.cs
gep13: Ooo! Missed that one! Will get that fixed up just now!

NGitLab/Models/PackageStatus.cs
gep13: Happy to make this change, but I am curious what your suggestion is when it comes to things like this:

https://github.com/ubisoft/NGitLab/pull/524/files#diff-27c124206cca88b65b78a9a2fe44d27dfcc3fab2b510dc692c6507c5cc7d8833R50

```
url = Utils.AddParameter(url, "status", query.Status);
```

i.e. when using the Enum value to construct the URL that is requested.  With the change that you have suggested in play, the constructed URL doesn't work, as the GitLab server returns an error...

image

Any suggestions?

gep13: @meziantou just wanted to follow up on this one.  Any ideas on the best course of action here? Thanks!

meziantou: Sorry for the delay. I think the best solution is to improve `Utils.AddParameter` to better support enums:

```
public static string AddParameter<T>(string url, string parameterName, T value)
{
    if (typeof(T).IsEnum)
    {
        // TODO
        // Note: ReflectionExtensions.GetEnumMappings can be used as an example
    }

    return Equals(value, null) ? url : AddParameterInternal(url, parameterName, value.ToString());
}
```

ap0llo:
Sorry for barging in, but I was actually looking into adding support for GitLab's package registry to NGitLab myself when I saw there is already an open PR.

Is there anything I can do to complete this? I'd be happy to help out

gep13:
I keep trying to get back to the PR, but I just haven't had the cycles to do it yet ☹️

From memory, all that was remaining was some re-work for the tests to not depend on file system directly. 

If you have some time, I have no objections to you creating a PR into my fork, and then I can update this PR.

ap0llo:
> If you have some time, I have no objections to you creating a PR into my fork, and then I can update this PR.

Great, I'll see what I can do

louis-z:
@gep13, @ap0llo, this PR is stale. Is there a plan to resurrect it any time soon? If not, we'll consider closing it.

ap0llo:
Yes, you're right.

I'm fine with closing it.

gep13:
@louis-z thank you for the follow up here! This is definitely something that is still on my todo list, its just that its the eternal problem where that todo list never gets any shorter, just longer!

PR #528, opened 2023-09-22
Support sub modules in mocks

jairbubbles:
NB:
- LibGit2Sharp does not support adding sub modules, using git command line instead (see https://github.com/libgit2/libgit2sharp/pull/1501)
- `-c protocol.file.allow=always` is needed as otherwise we get `git fatal: transport 'file' not allowed`

### DIFF
--- a/NGitLab.Mock/Config/GitLabHelpers.cs
+++ b/NGitLab.Mock/Config/GitLabHelpers.cs
@@ -449,6 +449,20 @@ namespace NGitLab.Mock.Config
         }
 
         /// <summary>
+        /// Add a submodule in commit
+        /// </summary>
+        public static GitLabCommit WithSubModule(this GitLabCommit commit, string projectName)
+        {
+            return Configure(commit, _ =>
+            {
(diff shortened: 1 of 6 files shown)


NGitLab.Mock/Config/GitLabHelpers.cs
```diff
@@ -1302,6 +1319,44 @@ private static Commit CreateCommit(GitLabServer server, Project prj, GitLabCommi
...
+                    throw new GitLabException($"Project {submodule.ProjectName} can't be found.");
+                }
+
+                if (!subModuleProject.Repository.GetCommits().Any())
+                {
+                    throw new GitLabException("Project added as a module must have least one commit.");
+                }
+
+                using var process = Process.Start(
+                    new ProcessStartInfo("git", $"-c protocol.file.allow=always submodule add {subModuleProject.SshUrl} {subModuleProject.Name}")
```
meziantou:
```suggestion
                    new ProcessStartInfo("git", $"-c protocol.file.allow=always submodule add \"{subModuleProject.SshUrl}\" \"{subModuleProject.Name}\"")
```

Issue #491, opened 2023-07-18
ID needs to change to long or string


IvanJosipovic:
The Job IDs are at "4681067651" in my GitLab.com instance. However, this may impact all IDs.

Running the following code throws an exception:
```
var projects = client.Projects.Get(new ProjectQuery() { });

foreach (var project in projects)
{
    var pipelines = client.GetPipelines(project.Id);

    foreach (var pipeline in pipelines.AllJobs)
    {

    }
}
```

```
System.Text.Json.JsonException
  HResult=0x80131500
  Message=The JSON value could not be converted to System.Int32. Path: $[0].id | LineNumber: 0 | BytePositionInLine: 17.
  Source=System.Text.Json
  StackTrace:
   at System.Text.Json.ThrowHelper.ReThrowWithPath(ReadStack& state, Utf8JsonReader& reader, Exception ex)
   at System.Text.Json.Serialization.JsonConverter`1.ReadCore(Utf8JsonReader& reader, JsonSerializerOptions options, ReadStack& state)
   at System.Text.Json.JsonSerializer.ReadFromSpan[TValue](ReadOnlySpan`1 utf8Json, JsonTypeInfo jsonTypeInfo, Nullable`1 actualByteCount)
   at System.Text.Json.JsonSerializer.ReadFromSpan[TValue](ReadOnlySpan`1 json, JsonTypeInfo jsonTypeInfo)
   at System.Text.Json.JsonSerializer.Deserialize[TValue](String json, JsonSerializerOptions options)
   at NGitLab.Impl.Json.Serializer.Deserialize[T](String json) in /_/NGitLab/Impl/Json/Serializer.cs:line 29
   at NGitLab.Impl.HttpRequestor.Enumerable`1.<GetEnumerator>d__5.MoveNext() in /_/NGitLab/Impl/HttpRequestor.cs:line 193
   at GitLabTool.Worker.<ExecuteAsync>d__4.MoveNext() in C:\Repos\GitLabTool\Worker.cs:line 31
   at Microsoft.Extensions.Hosting.Internal.Host.<StartAsync>d__12.MoveNext()
   at Microsoft.Extensions.Hosting.HostingAbstractionsHostExtensions.<RunAsync>d__4.MoveNext()
   at Microsoft.Extensions.Hosting.HostingAbstractionsHostExtensions.<RunAsync>d__4.MoveNext()
   at Microsoft.Extensions.Hosting.HostingAbstractionsHostExtensions.Run(IHost host)
[...]

PR #550, opened 2023-10-25
Support for using URL-encoded paths as group or project ids

PMExtra:
https://docs.gitlab.com/ee/api/rest/index.html#namespaced-path-encoding


NGitLab.Mock/Clients/ClusterClient.cs
```diff
@@ -14,6 +14,12 @@ public ClusterClient(ClientContext context, int projectId)
             _projectId = projectId;
         }
 
+        public ClusterClient(ClientContext context, string projectId)
```
Toa741: We indeed don't have the choice to create a new constructor everywhere for adding the support for project/group paths.

However, instead of just a string, it might be better to use a dedicated struct as a parameter that would contains either a long for the id (that would help for https://github.com/ubisoft/NGitLab/issues/491) or a string for the path.

```csharp
public readonly struct ProjectIdOrPath
{
    private readonly long _id;
    private readonly string? _path;

    public ProjectIdOrPath(long id) => _id = id;
    public ProjectIdOrPath(string path) => _path = path;

    public static implicit operator ProjectIdOrPath(long id) => new(id);
    public static implicit operator ProjectIdOrPath(string path) => new(path);

    internal readonly string ValueAsString => _path ?? _id.ToString(CultureInfo.InvariantCulture);
    internal readonly string ValueAsUriParameter => Uri.EscapeDataString(_path ?? _id.ToString(CultureInfo.InvariantCulture));

    public override string ToString() => ValueAsString;
}
```

- The new constructors will use `ProjectIdOrPath`. Implicit operators will make it simple to use.
- The old one should still work and at some point, can be marked as obsolete
- Wherever it's needed, we could use `ValueAsUriParameter` to construct URLs, to avoid doing that by hand every time
- We can have a similar `GroupIdOrPath`

meziantou: Note that you can add additional implicit converters if it makes the code easier to use. For instance,

````
public static implicit operator ProjectIdOrPath(Project project) => new(project.Id);
````

Comment (PMExtra) [reply NGitLab.Mock/Clients/ClusterClient.cs]: Awesome! It's a great design, I will follow it later.

Comment (PMExtra) [reply NGitLab.Mock/Clients/ClusterClient.cs]: There is totally no difference between `ProjectIdOrPath` and `GroupIdOrPath`.

I think we should use the same struct, or make a base struct at least.

Comment (meziantou) [reply NGitLab.Mock/Clients/ClusterClient.cs]: While there is no difference yet, it can change in the future. As I mentioned, you could add specific implicit converters to the struct.

Also, having a struct per concept allows to clearly see where a concept is used. It can also improve readability and avoid issues when a method accepts multiple similar parameters. For instance, `Get(Id groupId, Id projectId)` vs `Get(GroupId group, ProjectId project)`. I don't think there is a method with such signature yet, but it's more future proof.

If there are many similar structs in the future, we may use a code generator to write the boilerplate. I don't think we need this complexity at the moment as there are only 2 structs.

> make a base struct at least

Structs cannot be inherited

Comment (PMExtra) [reply NGitLab.Mock/Clients/ClusterClient.cs]: I think `groupId` will never be passed with `projectId` at a single request.

`projectId` is a unique identifier of a specific project. It contains the groupId internally if it's a namespaced path.

And in GitLab docs, there's a [**single** section describing namespaced path](https://docs.gitlab.com/ee/api/rest/index.html#namespaced-path-encoding). 
Both `projectId` and `groupId` are pointed to this section.

> Structs cannot be inherited

Can we use `record` or just `class`?

Comment (PMExtra) [reply NGitLab.Mock/Clients/ClusterClient.cs]: So, to sum up. I made a struct named `IdOrNamespacedPath` for both group and project for now.

(It's called `NamespacedPath` in GitLab docs, but `PathWithNamespace` may be clearer. What are your thoughts on this?)

**Comment by PMExtra** [comment]
I changed the existing constructor to only support the new `ProjectId` or `GroupId` if the class is `internal`. Else, I marked the old constructor be obsolete and made a new one.

Maybe we should unify the client implementations classes to be `internal` instead of `public`? But it's a breaking change, although most users will never create a client by manually calling the constructor.

File: NGitLab.Mock/Clients/ClusterClient.cs (hunk 1 above)
Comment (PMExtra) [reply NGitLab.Mock/Clients/ClusterClient.cs]: Finally, I made 2 records type `ProjectId` and `GroupId` that inherits from `record IdOrNamespacedPath`.

File: NGitLab/Impl/ProjectBadgeClient.cs [hunk 2]
```diff
@@ -1,14 +1,13 @@
...
-using NGitLab.Extensions;
 using NGitLab.Models;
 
 namespace NGitLab.Impl
 {
     internal sealed class ProjectBadgeClient : BadgeClient, IProjectBadgeClient
     {
-        public ProjectBadgeClient(API api, int projectId)
-            : base(api, Project.Url + $"/{projectId.ToStringInvariant()}")
+        public ProjectBadgeClient(API api, ProjectId projectId)
```
Toa741, line 9: Breaking change, is it normal that it doesn't have a new constructor like the other ones?

NGitLab.Mock/Clients/ClusterClient.cs
Toa741: - I would tend to say that struct is still a valid option, even if it's duplicated. The project at the moment don't make an heavy use of either struct or records.
- I would have thought that the project already has nullable enabled ^^. Though it doesn't seem to be the case.

@louis-z , @meziantou ,
- Is the use of record in this case a valid option? Should we enforce a choice on tiny value types as those IDs?
- I suppose that not having nullable enabled on project might be because of the amount of work to do so. Though, is having `#nullable enable` acceptable on new files?

NGitLab/Impl/ProjectBadgeClient.cs
PMExtra: It's an internal sealed class

NGitLab/Impl/MilestoneClient.cs
```diff
@@ -12,10 +12,16 @@ public class MilestoneClient : IMilestoneClient
...
+        {
+        }
+
+        internal MilestoneClient(API api, MilestoneScope scope, IdOrNamespacedPath projectId)
         {
             _api = api;
-            _milestonePath = $"/{scope.ToString().ToLowerInvariant()}/{id.ToStringInvariant()}/milestones";
+            _milestonePath = $"/{scope.ToString().ToLowerInvariant()}/{projectId.ValueAsUriParameter}/milestones";
             Scope = scope;
         }
```
PMExtra: @Toa741 A base type of `ProjectId` and `GroupId` will make some methods reusable.

NGitLab.Mock/Clients/ClusterClient.cs
meziantou: > Is the use of record in this case a valid option? Should we enforce a choice on tiny value types as those IDs?

You'll need to handle null values everywhere. If the goal is to share some behavior, you can also use an interface (IIdOrPathAddressable).

> I suppose that not having nullable enabled on project might be because of the amount of work to do so. Though, is having #nullable enable acceptable on new files?

Historic reasons, but it would be nice to use Nullable Reference Types at least on new code.

PMExtra: @Toa741 @meziantou I met a problem while use `#nullable enable` for a single file:

```
RS0037: PublicAPI.txt is missing '#nullable enable', so the nullability annotations of API isn't recorded. It is recommended to enable this tracking.
```

If I put `#nullable enable` to `PublicAPI.txt`, all other API without Nullable Reference will be broken:

```
RS0036: Symbol '***' is missing nullability annotations in the declared API
RS0041: Symbol '***' uses some oblivious reference types
```

Toa741: - True for handling null everywhere technically. @PMExtra could we give the struct + interface a try?
- For the nullable enable in PublicAPI, maybe that's another reason why we didn't started to mark new files as nullable. Not sure there would be a way around that

PMExtra:
@Toa741 I think it's ready to merge. Please let me know if there is anything else.

Toa741:
> @Toa741 I think it's ready to merge. Please let me know if there is anything else.

Alright :), will make a final review

NGitLab/Models/IidOrPathAddressable.cs
```diff
@@ -0,0 +1,11 @@
+#nullable enable
+
+namespace NGitLab.Models
+{
+    public interface IidOrPathAddressable
```
Toa741:
```suggestion
    public interface IIdOrPathAddressable
```

PR #562, opened 2023-11-15
Add runner registration/delete mock

BallyB:
We implemented the Register/Delete logic for the runner in the NGitLab Mock project. We needed these to unit test some use cases of runner registration.

We added the token registration properties in the project even though we are aware that it will be soon removed from the API : https://docs.gitlab.com/runner/register/#register-with-a-runner-registration-token-deprecated

### DIFF
--- a/NGitLab.Mock/Clients/RunnerClient.cs
+++ b/NGitLab.Mock/Clients/RunnerClient.cs
@@ -71,7 +71,7 @@ namespace NGitLab.Mock.Clients
 
                 if (projects.Take(2).Count() > 1)
                 {
-                    throw new GitLabException("Bad Request. Runner is enabled in multiple projects");
+                    throw new GitLabBadRequestException("Runner is enabled in multiple projects");
                 }
 
                 var project = GetProject(projects.Single().Id, ProjectPermission.Edit);
--- a/NGitLab.Mock/Clients/RunnerClient.cs
+++ b/NGitLab.Mock/Clients/RunnerClient.cs
@@ -69,7 +69,7 @@ namespace NGitLab.Mock.Clients
                     throw new GitLabBadRequestException("Runner is not found in any project");
                 }
 
-                if (projects.Take(2).Count() > 1)
+                if (projects.Skip(1).Any())
                 {
                     throw new GitLabBadRequestException("Runner is enabled in multiple projects");
                 }
@@ -142,7 +142,7 @@ namespace NGitLab.Mock.Clients
 
                 if (project.EnabledRunners.Contains(runnerReference))
                 {
-                    throw new GitLabException("Bad Request. Runner has already been taken");
+                    throw new GitLabBadRequestException("Runner has already been taken");
                 }
 
                 project.EnabledRunners.Add(runnerReference);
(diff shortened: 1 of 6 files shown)


NGitLab.Mock/GitLabServer.cs
```diff
@@ -130,11 +132,25 @@ public void Dispose()
...
         }
 
+        internal string MakeToken(string id, string prefix = "")
+        {
+            return prefix + id.PadLeft(20, '0');
+        }
+
+        internal string MakeRegistrationToken(string id)
+        {
+            return MakeToken(id, "GR1348941");
```
meziantou, line 152:
```suggestion
            return MakeToken(id, "GR1234567A");
```

NGitLab.Mock/Project.cs
```diff
@@ -149,6 +149,8 @@ public string[] Tags
 
         public ProtectedBranchCollection ProtectedBranches { get; }
 
+        public string RunnersToken { get; set; }
```
meziantou: I don't think you need a setter. The value should be generated in the ctor and I don't think it can be changed.

If you actually need a setter, maybe it could be `internal`

```suggestion
        public string RunnersToken { get; }
```

NGitLab.Mock/Clients/RunnerClient.cs
```diff
@@ -57,14 +57,26 @@ public RunnerClient(ClientContext context)
...
 
         public void Delete(int runnerId)
         {
-            throw new NotImplementedException();
+            using (Context.BeginOperationScope())
+            {
+                var projects = Server.AllProjects.Where(p => p.EnabledRunners.Any(r => r.Id == runnerId));
+                if (!projects.Any())
+                {
+                    throw new GitLabException("Bad Request. Runner is not found in any project");
```
meziantou:
```suggestion
                    throw new GitLabBadRequestException("Runner is not found in any project");
```

NGitLab.Mock/Clients/RunnerClient.cs
```diff
@@ -57,14 +57,26 @@ public RunnerClient(ClientContext context)
...
+                }
+
+                if (projects.Take(2).Count() > 1)
+                {
+                    throw new GitLabException("Bad Request. Runner is enabled in multiple projects");
```
meziantou:
```suggestion
                    throw new GitLabBadRequestException("Runner is enabled in multiple projects");
```

meziantou (changes requested):

NGitLab.Mock/Project.cs
BallyB: I need a setter because when you instantiate a new project, the server is still null. If we want to keep the token generation logic on server side, we have to generate the project's token when we link it to the group/server collection

NGitLab.Mock/GitLabServer.cs
BallyB: The prefix I used is the real prefix used by GitLab for every registration token. I think it can help to directly recognize it

BallyB: https://gitlab.com/gitlab-org/gitlab/-/issues/388379

Issue #563, opened 2023-11-17
Missing get job artifact by ref and filepath


shivan:
I miss the api request to get job artifacts by ref and file path.

see
https://docs.gitlab.com/ee/api/job_artifacts.html#download-a-single-artifact-file-from-specific-tag-or-branch

PR #564, opened 2023-11-17
Add ability to get job artifact by Ref

shivan:
see #563

### DIFF
--- a/NGitLab/Models/JobArtifactQuery.cs
+++ b/NGitLab/Models/JobArtifactQuery.cs
@@ -2,7 +2,7 @@
 
 namespace NGitLab.Models
 {
-    public class JobArtifactQuery
+    public sealed class JobArtifactQuery
     {
         [JsonPropertyName("ref_name")]
         public string RefName;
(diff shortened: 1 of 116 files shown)


NGitLab/Models/JobArtifactQuery.cs
```diff
@@ -0,0 +1,16 @@
...
+    public class JobArtifactQuery
+    {
+        [JsonPropertyName("ref_name")]
+        public string RefName;
+
+        [JsonPropertyName("artifact_path")]
+        public string ArtifactPath;
+
+        [JsonPropertyName("job")]
+        public string JobName;
```
louis-z: Please define properties rather than fields.

louis-z (changes requested):

NGitLab/Models/JobArtifactQuery.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System.Text.Json.Serialization;
+
+namespace NGitLab.Models
+{
+    public class JobArtifactQuery
```
louis-z:
```suggestion
    public sealed class JobArtifactQuery
```

NGitLab/Models/JobArtifactQuery.cs
shivan: ok

shivan:
@louis-z any further changes requested? Or all ok now?

louis-z:
> @louis-z any further changes requested? Or all ok now?

I had a quick look and it seems fine. Can you merge or rebase from the main branch? Note that there are conflicts. I'll have one more look at it tomorrow.

PR #577, opened 2023-11-23
Add protected tags client

Sjhzjxc:
https://docs.gitlab.com/ee/api/protected_tags.html

### DIFF
--- a/NGitLab.Tests/ProtectedTagTests.cs
+++ b/NGitLab.Tests/ProtectedTagTests.cs
@@ -1,0 +1,61 @@
+using System.Linq;
+using System.Threading.Tasks;
+using NGitLab.Models;
+using NGitLab.Tests.Docker;
+using NUnit.Framework;
+
+namespace NGitLab.Tests
+{
+    public class ProtectedTagTests
+    {
+        [Test]
+        [NGitLabRetry]
+        public async Task ProtectTag_Test()
+        {
+            using var context = await GitLabTestContext.CreateAsync();
+            var project = context.CreateProject(initializeWithCommits: true);
+            var tagClient = context.Client.GetRepository(project.Id).Tags;
+
+            var tag = tagClient.Create(new TagCreate { Name = "protectedTag", Ref = project.DefaultBranch });
+            Assert.False(tag.Protected);
+
+            var protectedTagClient = context.Client.GetProtectedTagClient(project.Id);
+            var protectedTags = protectedTagClient.GetProtectedTagsAsync();
+            Assert.False(protectedTags.Any());
+
+            await ProtectedTagTest(tag.Name, tag.Name, protectedTagClient, tagClient);
+            await ProtectedTagTest("*", tag.Name, protectedTagClient, tagClient);
+        }
+
+        private static async Task ProtectedTagTest(string tagProtectName, string tagName, IProtectedTagClient protectedTagClient, ITagClient tagClient)
+        {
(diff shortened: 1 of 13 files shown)


Sjhzjxc:
https://docs.gitlab.com/ee/api/protected_branches.html#protect-repository-branches

NGitLab/Models/BranchProtect.cs
```diff
@@ -25,13 +25,13 @@ public BranchProtect(string branchName)
         public bool AllowForcePush { get; set; } = false;
 
         [JsonPropertyName("allowed_to_merge")]
-        public AccessLevelInfo[] AllowedToMerge { get; set; }
+        public AccessControl[] AllowedToMerge { get; set; }
```
Toa741: Although it make sense to have AccessControl there, it's also a breaking change for BranchProtect. Is this change mandatory for you? Otherwise, I would suggest to keep AccessLevelInfo on BranchProtect at the moment and push ProtectedTag with AccessControl as it's new.

Then, there are multiple fixes we would like to make and also requires breaking changes like https://github.com/ubisoft/NGitLab/issues/491. We could add the AccessControl update in the future rework.

NGitLab.Tests/ProtectedTagTests.cs
```diff
@@ -0,0 +1,61 @@
...
+        private static async Task ProtectedTagTest(string tagProtectName, string tagName, IProtectedTagClient protectedTagClient, ITagClient tagClient)
+        {
+            var tagProtect = new TagProtect(tagProtectName)
+            {
+                AllowedToCreate = new AccessControl[] { new AccessLevelControl { AccessLevel = AccessLevel.Maintainer }, },
+                CreateAccessLevel = AccessLevel.NoAccess,
+            };
+            var protectedTag = await protectedTagClient.ProtectTagAsync(tagProtect).ConfigureAwait(false);
+            Assert.AreEqual(protectedTag.Name, tagProtect.Name);
+            var accessList = protectedTag.CreateAccessLevels.Select(level => level.AccessLevel).ToList();
```
Toa741:
```suggestion
            var accessList = protectedTag.CreateAccessLevels.Select(level => level.AccessLevel).ToArray();
```

NGitLab.Tests/ProtectedTagTests.cs
```diff
@@ -0,0 +1,61 @@
...
+            Assert.Contains(AccessLevel.NoAccess, accessList);
+            Assert.Contains(AccessLevel.Maintainer, accessList);
+
+            var getProtectedTag = await protectedTagClient.GetProtectedTagAsync(tagProtectName).ConfigureAwait(false);
+            Assert.AreEqual(protectedTag.Name, getProtectedTag.Name);
+            accessList = getProtectedTag.CreateAccessLevels.Select(level => level.AccessLevel).ToList();
```
Toa741:
```suggestion
            accessList = getProtectedTag.CreateAccessLevels.Select(level => level.AccessLevel).ToArray();
```

NGitLab/Models/BranchProtect.cs
Sjhzjxc: This change isn't mandatory for me. I modified `BranchProtect` because introducing the `Id` attribute into `AccessLevelInfo` caused an error in `ProtectBranch_Test`.

I have removed the `Id` attribute from `AccessLevelInfo` and revoked the changes to `BranchProtect`.

PR #591, opened 2023-12-05
Add support for project's "releases_access_level" + a bit more

louis-z:


NGitLab.Tests/ReleaseClientTests.cs
```diff
@@ -30,6 +31,9 @@ public async Task Test_release_api()
                 Description = "test",
             });
 
+            if (context.IsGitLabVersionInRange(VersionRange.Parse("[15.5,)"), out _))
```
meziantou: Is the check needed?

meziantou (approved):

NGitLab.Tests/ReleaseClientTests.cs
louis-z: Probably not. Removed.

PR #620, opened 2024-01-25
Get Group MergeRequests

Yekonori:


NGitLab.Mock.Tests/MergeRequestsMockTests.cs
```diff
@@ -412,4 +412,28 @@ public void Test_merge_request_resource_milestone_events_found()
...
+            .WithProject("project1", @namespace: "parentGroup", configure: project => project
+                .WithMergeRequest("branch-01", title: "Merge request 1", author: "user1", assignee: "user2")
+                .WithMergeRequest("branch-02", title: "Merge request 2", author: "user2", assignee: "user1"))
+            .WithProject("project2", @namespace: "parentGroup", configure: project => project
+                .WithMergeRequest("branch-03", title: "Merge request 3", author: "user1", assignee: "user2")
+                .WithMergeRequest("branch-04", title: "Merge request 4", author: "user2", assignee: "user1"))
+            .BuildServer();
+
+        var client = server.CreateClient("user1");
+        var mrClient = client.GetGroupMergeRequest(groupId);
```
Toa741: I think this is working by luck:

- In the mocked Client, you set the `_groupid` but it is used nowhere
- Probably that the mocked client still rely on _projectid, and fallback to all projects/MR as it is null
- Thus, if you introduce in your test another group that has another project with MR, the test will fail (as it will get the MR for the other group).

Ideally the mocked client should handle both case (project or group). Worst case I would not make the test and throw an not implemented/supported exception on the group constructor (unless you actually need a working mock for this scenario :)).

NGitLab/IGitLabClient.cs
```diff
@@ -79,6 +79,11 @@ public interface IGitLabClient
 
     IMergeRequestClient GetMergeRequest(ProjectId projectId);
 
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    IMergeRequestClient GetGroupMergeRequest(int groupId);
```
louis-z: Why have this method with the `[EditorBrowsable(EditorBrowsableState.Never)]` attribute here?

If we should avoid calling the method, then it simply should not be added to the interface.

BTW the only reason we have the attribute on, say, the `IMilestoneClient GetMilestone(int projectId)` declaration is because that method was already part of the public API when we decided to make it "obsolete".

PR #621, opened 2024-01-25
Add position property when getting merge request comments

Codex04:
Related api endpoint: https://docs.gitlab.com/ee/api/discussions.html#list-project-merge-request-discussion-items

### DIFF
--- a/NGitLab/Models/Position.cs
+++ b/NGitLab/Models/Position.cs
@@ -4,17 +4,20 @@ namespace NGitLab.Models;
 
 public class Position
 {
+    [JsonPropertyName("old_path")]
+    public string OldPath { get; set; }
+
     [JsonPropertyName("new_path")]
     public string NewPath { get; set; }
 
+    [JsonPropertyName("position_type")]
+    public DynamicEnum<PositionType> PositionType { get; set; }
+
     [JsonPropertyName("old_line")]
-    public int OldLine { get; set; }
+    public int? OldLine { get; set; }
 
     [JsonPropertyName("new_line")]
-    public int NewLine { get; set; }
-
-    [JsonPropertyName("position_type")]
-    public string PositionType { get; set; }
+    public int? NewLine { get; set; }
 
     [JsonPropertyName("line_range")]
     public LineRange LineRange { get; set; }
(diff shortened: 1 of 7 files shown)


NGitLab/Models/Position.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace NGitLab.Models;
+
+public class Position
+{
+    [JsonPropertyName("new_path")]
+    public string NewPath { get; set; }
+
+    [JsonPropertyName("old_line")]
+    public int OldLine { get; set; }
```
louis-z: It should be nullable and so should `NewLine`.

NGitLab/Models/Position.cs
```diff
@@ -0,0 +1,21 @@
+using System.Text.Json.Serialization;
+
+namespace NGitLab.Models;
+
+public class Position
+{
+    [JsonPropertyName("new_path")]
+    public string NewPath { get; set; }
```
louis-z, line 11: Shouldn't we also decode "old_path"?

NGitLab/Models/Position.cs
```diff
@@ -0,0 +1,21 @@
...
+
+    [JsonPropertyName("old_line")]
+    public int OldLine { get; set; }
+
+    [JsonPropertyName("new_line")]
+    public int NewLine { get; set; }
+
+    [JsonPropertyName("position_type")]
+    public string PositionType { get; set; }
```
louis-z: Can you introduce an enum for this? Currently allowed values seem to be `text` or `image`, but to make this more robust and support any other eventual types. you could wrap the enum into a `DynamicEnum<TEnum>`.

NGitLab/Models/Range.cs
```diff
@@ -0,0 +1,12 @@
...
+
+namespace NGitLab.Models;
+
+public class Range
+{
+    [JsonPropertyName("line_code")]
+    public string LineCode { get; set; }
+
+    [JsonPropertyName("type")]
+    public string Type { get; set; }
```
louis-z: I think this too could be represented by an enum, with possible values: `old` and `new`.

louis-z (changes requested):

louis-z (approved):

PR #641, opened 2024-02-29
Add CreateAsync to UserClient

kevingentile:
Adds `CreateAsync` to `IUserClient`

### DIFF
--- a/NGitLab.Mock/Clients/UserClient.cs
+++ b/NGitLab.Mock/Clients/UserClient.cs
@@ -64,6 +64,23 @@ internal sealed class UserClient : ClientBase, IUserClient
         }
     }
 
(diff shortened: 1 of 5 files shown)


NGitLab.Mock/Clients/UserClient.cs
```diff
@@ -64,6 +64,23 @@ public Models.User Create(UserUpsert user)
...
+            {
+                Name = user.Name,
+                Email = user.Email,
+            };
+
+            Server.Users.Add(u);
+
+            await Task.Yield();
+            return u.ToClientUser();
+        }
```
belcher-rok:
```suggestion
        await Task.Yield();
        return Create(user);
```

belcher-rok (changes requested):

belcher-rok (approved):

PR #642, opened 2024-03-04
Update `IGroupsClient` to support paging of groups and projects

belcher-rok:
Added ability to enumerate groups, subgroups, and projects by page number. Each successful call returns a collection of items in addition to the total number of items.

All paging methods start with "Page", all are async, and they use `GroupId` to id the target group where applicable.

Also
- Added ability to query top-level groups (see `GroupQuery.TopLevelOnly`).
- Added ability to query descendant subgroups (see `SubgroupQuery.IncludeDescendants`).
- Added ability for `IHttpRequestor` to access the response headers if necessary.
- Added equality operators to `GroupId` and `ProjectId`.
- Added overloads taking a `GroupId` as long as doing so would not result in ambiguous calls for existing methods.
- All changes should be backward compatible, at least that is the intention.


NGitLab/Models/GroupId.cs
```diff
@@ -13,24 +24,89 @@ namespace NGitLab.Models;
...
-        _id = id;
+        _id = id != 0 ? id : throw new ArgumentException("Id cannot be zero", nameof(id));
     }
 
     public GroupId(string path)
     {
         _path = path ?? throw new ArgumentNullException(nameof(path));
     }
 
-    public GroupId(Group group)
```
belcher-rok, line 24: Removed this c'tor because the resulting `GroupId` is ambiguous. For example,
```C#
Group group = GetGroup(...);
var id1 = new GroupId(group); // calls the removed c`tor
var id2 = new GroupId(group.Id);
var id3 = new GroupId(group.FullPath);

Assert(id1 == id2);  // Ok, both contain `group.Id`
Assert(id1 == id3);  // Fails! group.Id != group.FullPath.
Assert(id2 == id3);  // Fails for same reason, but there no easy way around this other than to not do it.

// Preferred usage:
Assert(id1 == group); // Ok!
Assert(id2 == group); // Ok!
Assert(id3 == group); // Ok!
```

belcher-rok:
@louis-z Do you need anything else from me?

louis-z:
> @louis-z Do you need anything else from me?

I haven't had the chance to look at this PR yet. Will let you know...

NGitLab/IHttpRequestor.cs
```diff
@@ -12,10 +13,18 @@ public interface IHttpRequestor
...
+
+    Task<(IReadOnlyCollection<T> Page, int? Total)> PageAsync<T>(string tailAPIUrl, CancellationToken cancellationToken);
+
     void Stream(string tailAPIUrl, Action<Stream> parser);
 
+    void StreamAndHeaders(string tailAPIUrl, Action<Stream, WebHeaderCollection> parser);
+
     Task StreamAsync(string tailAPIUrl, Func<Stream, Task> parser, CancellationToken cancellationToken);
 
+    Task StreamAndHeadersAsync(string tailAPIUrl, Func<Stream, WebHeaderCollection, Task> parser, CancellationToken cancellationToken);
```
louis-z: Do not expose in the public API any of `System.Net`'s types. We have plans to eventually steer away from the `Web*` classes (e.g. `WebRequest`) completely and use `HttpClient` instead.

Maybe you could expose headers via a `IReadOnlyDictionary<string, string>` instead?

NGitLab/IGroupsClient.cs
```diff
@@ -9,53 +9,183 @@ namespace NGitLab;
...
+    /// <inheritdoc cref="SearchProjectsAsync"/>
+    /// <param name="id">The group's id</param>
+    /// <param name="search">The search criteria.</param>
+    [Obsolete("Use SearchProjectsAsync instead")]
+    IEnumerable<Project> SearchProjects(int id, string search);
 
-    GitLabCollectionResponse<Project> GetProjectsAsync(int groupId, GroupProjectsQuery query);
+    /// <inheritdoc cref="SearchProjectsAsync"/>
+    /// <param name="id">The group's id</param>
+    GitLabCollectionResponse<Project> GetProjectsAsync(int id, GroupProjectsQuery query);
```
louis-z: Don't rename `groupId` to `id`, as this is a breaking change.

NGitLab/IGroupsClient.cs
```diff
@@ -9,53 +9,183 @@ namespace NGitLab;
...
+    /// <returns>The new group.</returns>
     Task<Group> CreateAsync(GroupCreate group, CancellationToken cancellationToken = default);
 
-    [Obsolete("Use GetProjectsAsync instead")]
-    IEnumerable<Project> SearchProjects(int groupId, string search);
+    /// <inheritdoc cref="SearchProjectsAsync"/>
+    /// <param name="id">The group's id</param>
+    /// <param name="search">The search criteria.</param>
+    [Obsolete("Use SearchProjectsAsync instead")]
+    IEnumerable<Project> SearchProjects(int id, string search);
```
louis-z: Don't rename `groupId` to `id`, as this is a breaking change.

NGitLab/Models/GroupId.cs
louis-z: It's probably wise to eventually get rid of this ctor, but for now please leave it in and tag is with the `[Obsolete]` attribute, so as not to introduce a breaking change.

NGitLab/Models/ProjectId.cs
```diff
@@ -13,24 +24,89 @@ namespace NGitLab.Models;
...
-        _id = id;
+        _id = id != 0 ? id : throw new ArgumentException("Id cannot be zero", nameof(id));
     }
 
     public ProjectId(string path)
     {
         _path = path ?? throw new ArgumentNullException(nameof(path));
     }
 
-    public ProjectId(Project project)
```
louis-z, line 24: Please leave this ctor in and simply tag it `[Obsolete]` for now.

louis-z (changes requested):

louis-z:
Hey @belcher-rok, thanks a lot for your contribution! This is great work!

Adding/improving comments was a nice touch too, but when documenting a method, could you please use the 3rd person (*descriptive form*, such as `gets`) rather than the 2nd person (*prescriptive form*, such as `get`)? This is a widespread standard which, unfortunately, we did not always comply with in the past.

Check out
- [The Google doc guidelines](https://developers.google.com/style/reference-verbs)
- [An example from Microsoft doc](https://docs.microsoft.com/en-us/dotnet/api/system.io.file.getattributes?view=net-6.0#definition)

NGitLab.Mock.Tests/GroupsMockTests.cs
```diff
@@ -78,4 +151,192 @@ public void Test_get_subgroups_by_fullpath()
...
+
+    [Test]
+    public void Test_page_subgroups_with_invalid_perpage_throws()
+    {
+        using var server = CreateGroupHierarchy();
+        var client = server.CreateClient("user1");
+        Assert.ThrowsAsync<Clients.GitLabBadRequestException>(() => client.Groups.PageSubgroupsAsync(1, new(page: 1, perPage: 0)));
+    }
+
+    private static GitLabServer CreateProjectHierarchy() =>
```
Toa741: Can it be moved at the top of the file instead? Even though I understand it was to keep it close to the related tests, let's keep static fields together.

NGitLab.Mock/Clients/GroupClient.cs
```diff
@@ -155,28 +122,66 @@ public IEnumerable<Models.Group> Get(GroupQuery query)
...
+    public async Task<(IReadOnlyCollection<Models.Group> Page, int? Total)> PageAsync(PageQuery<GroupQuery> query, CancellationToken cancellationToken = default)
     {
-        return GitLabCollectionResponse.Create(Get(query));
+        await Task.Yield();
+        var pageNum = Math.Max(1, query?.Page ?? PageQuery.FirstPage);
+        var perPage = query?.PerPage ?? PageQuery.DefaultPerPage;
+        if (perPage < PageQuery.MinPerPage)
+        {
+            // Max isn't enforced the same way
+            throw new GitLabBadRequestException("per_page has a value not allowed");
```
Toa741:
```suggestion
            throw new GitLabBadRequestException($"per_page value ({perPage}) is invalid: cannot be lower than MinPerPage ({PageQuery.MinPerPage})");
```

To give more context.

NGitLab.Mock/Clients/GroupClient.cs
```diff
@@ -155,28 +122,66 @@ public IEnumerable<Models.Group> Get(GroupQuery query)
...
+        }
+
+        var all = Get(query?.Query).ToList();
```
Toa741:
```suggestion
        var all = Get(query?.Query).ToArray();
```

NGitLab.Mock/Clients/GroupClient.cs
```diff
@@ -155,28 +122,66 @@ public IEnumerable<Models.Group> Get(GroupQuery query)
...
+        var page = all
+            .Skip((pageNum - 1) * perPage)
+            .Take(perPage)
+            .ToList();
```
Toa741:
```suggestion
            .ToArray();
```

NGitLab.Mock/Clients/GroupClient.cs
```diff
@@ -231,6 +240,24 @@ public GitLabCollectionResponse<Models.Project> GetProjectsAsync(int groupId, Gr
...
     }
 
+    public async Task<(IReadOnlyCollection<Models.Project> Page, int? Total)> PageProjectsAsync(GroupId groupId, PageQuery<GroupProjectsQuery> query, CancellationToken cancellationToken = default)
+    {
+        await Task.Yield();
+        var pageNum = Math.Max(1, query?.Page ?? PageQuery.FirstPage);
+        var perPage = query?.PerPage ?? PageQuery.DefaultPerPage;
+        if (perPage < 1)
+        {
+            throw new GitLabBadRequestException("per_page has a value not allowed");
```
Toa741:
```suggestion
            throw new GitLabBadRequestException("per_page value ({perPage}) is invalid: cannot be inferior to 1");
```

NGitLab.Mock/Clients/GroupClient.cs
```diff
@@ -231,6 +240,24 @@ public GitLabCollectionResponse<Models.Project> GetProjectsAsync(int groupId, Gr
...
+        }
+
+        var all = SearchProjectsAsync(groupId, query?.Query).ToList();
```
Toa741:
```suggestion
        var all = SearchProjectsAsync(groupId, query?.Query).ToArray();
```

NGitLab.Mock/Clients/GroupClient.cs
```diff
@@ -231,6 +240,24 @@ public GitLabCollectionResponse<Models.Project> GetProjectsAsync(int groupId, Gr
...
+        var page = all
+            .Skip((pageNum - 1) * perPage)
+            .Take(perPage)
+            .ToList();
```
Toa741:
```suggestion
            .ToArray();
```

NGitLab.Mock/Clients/GroupClient.cs
```diff
@@ -320,42 +352,49 @@ public GitLabCollectionResponse<Models.Group> GetSubgroupsByIdAsync(int id, Subg
...
 
-            return GitLabCollectionResponse.Create(clientGroups.Where(g => g.ParentId == parentGroup.Id));
+    public async Task<(IReadOnlyCollection<Models.Group> Page, int? Total)> PageSubgroupsAsync(GroupId groupId, PageQuery<SubgroupQuery> query, CancellationToken cancellationToken = default)
+    {
+        await Task.Yield();
+        var pageNum = Math.Max(1, query?.Page ?? PageQuery.FirstPage);
+        var perPage = query?.PerPage ?? PageQuery.DefaultPerPage;
+        if (perPage < 1)
+        {
+            throw new GitLabBadRequestException("per_page has a value not allowed");
```
Toa741:
```suggestion
            throw new GitLabBadRequestException("per_page value ({perPage}) is invalid: cannot be inferior to 1");
```

NGitLab.Mock/Clients/GroupClient.cs
```diff
@@ -320,42 +352,49 @@ public GitLabCollectionResponse<Models.Group> GetSubgroupsByIdAsync(int id, Subg
...
         }
+
+        var all = GetSubgroupsAsync(groupId, query?.Query).ToList();
```
Toa741:
```suggestion
        var all = GetSubgroupsAsync(groupId, query?.Query).ToArray();
```

NGitLab.Mock/Clients/GroupClient.cs
```diff
@@ -320,42 +352,49 @@ public GitLabCollectionResponse<Models.Group> GetSubgroupsByIdAsync(int id, Subg
...
+        var page = all
+            .Skip((pageNum - 1) * perPage)
+            .Take(perPage)
+            .ToList();
```
Toa741:
```suggestion
            .ToArray();
```

NGitLab/IGroupsClient.cs
```diff
@@ -9,53 +9,183 @@ namespace NGitLab;
...
     /// </summary>
-    /// <remarks>https://github.com/gitlabhq/gitlabhq/blob/master/doc/api/groups.md#details-of-a-group</remarks>
+    /// <param name="query">The query parameters, including the page.</param>
+    /// <param name="cancellationToken">The cancellation token</param>
+    /// <returns>A page of groups and the total number of matching groups (used to compute the last page).</returns>
+    /// <remarks>
+    /// The page will be empty after the last page is read.<para/>
+    /// The total maybe null if there are more than 10,000 items. See https://docs.gitlab.com/ee/user/gitlab_com/index.html#pagination-response-headers.
+    /// </remarks>
+    Task<(IReadOnlyCollection<Group> Page, int? Total)> PageAsync(PageQuery<GroupQuery> query, CancellationToken cancellationToken = default);
```
Toa741: Not a huge fan of exposed Tuples. It looks a bit cumbersome as it's repeated in multiple places. Although it might not happen, if we need to add a property in the future it makes a lot of place to change too. 

Could we create a simple class for it (or a record)? `PaginatedResponse<T>` or something equivalent

NGitLab.Mock/Clients/GroupClient.cs
```diff
@@ -320,42 +352,49 @@ public GitLabCollectionResponse<Models.Group> GetSubgroupsByIdAsync(int id, Subg
...
+        return (page, all.Count > 10000 ? null : all.Count);
```
Toa741: Can the 10000 limit be added as a const?

NGitLab/Impl/GroupsClient.cs
```diff
@@ -36,61 +35,79 @@ public GitLabCollectionResponse<Group> GetAsync(GroupQuery query)
...
+    }
+
+    private static string CreateGetUrl(GroupQuery query, int? page = null, int? perPage = null)
+    {
+        var url = Url;
+
+        url = Utils.AddPageParams(url, page, perPage);
 
-        if (query.SkipGroups != null && query.SkipGroups.Any())
+        if (query is not null)
```
Toa741: You can invert the condition and return early. That will remove one level of indentation

NGitLab/Impl/HttpRequestor.cs
```diff
@@ -139,6 +173,14 @@ public virtual GitLabCollectionResponse<T> GetAllAsync<T>(string tailUrl)
         return new Enumerable<T>(_apiToken, GetAPIUrl(tailUrl), _options);
     }
 
+    private static string ReadText(Stream s) => new StreamReader(s).ReadToEnd();
```
Toa741: StreamReader is disposable

NGitLab.Mock/Clients/GroupClient.cs
belcher-rok: @Toa741 I don't have a problem making this change, I was just attempting to mimic the [actual GitLab error message](https://gitlab.com/gitlab-org/gitlab/-/blob/master/spec/requests/api/commits_spec.rb#L356). Should I still use a more descriptive message?

NGitLab/IHttpRequestor.cs
belcher-rok: Switched to `IReadOnlyDictionary<string, IEnumerable<string>>`. I had to create an adaptor class to do it. See `WebHeadersDictionaryAdaptor`.

NGitLab/IGroupsClient.cs
belcher-rok: Good idea. Created `PagedResponse<T>` type.

belcher-rok:
> Hey @belcher-rok, thanks a lot for your contribution! This is great work!
> 
> Adding/improving comments was a nice touch too, but when documenting a method, could you please use the 3rd person (_descriptive form_, such as `gets`) rather than the 2nd person (_prescriptive form_, such as `get`)? This is a widespread standard which, unfortunately, we did not always comply with in the past.
> 
> Check out
> 
> * [The Google doc guidelines](https://developers.google.com/style/reference-verbs)
> * [An example from Microsoft doc](https://docs.microsoft.com/en-us/dotnet/api/system.io.file.getattributes?view=net-6.0#definition)

@louis-z and @Toa741 Thanks for your reviews! You made the code better. I think I addressed all your suggestions except for one (see [above](https://github.com/ubisoft/NGitLab/pull/642#discussion_r1515294016)). Please let me know if you have any further requests.

A little background about me: My company is starting a project which utilizes a private GitLab instance. We are planning to use NGitLab as the client interface. It is quite likely that I will have additional PRs in the near future.

belcher-rok:
@louis-z and @Toa741 please let me know what I can do to help move this along.

NGitLab.Mock/Clients/GroupClient.cs
Toa741: Yes, the mock is primarily used for tests and I would be happier with a message that points the problem immediately, rather than trying to find the response in the doc.

Ideally, the API response would be more precise too ^^, `per_page has a value not allowed` is a bit vague

louis-z:
> @louis-z and @Toa741 please let me know what I can do to help move this along.

Please address the remaining conversations and update your branch WRT main. Once it's done, this PR should get merged.

belcher-rok:
> > @louis-z and @Toa741 please let me know what I can do to help move this along.
> 
> Please address the remaining conversations and update your branch WRT main. Once it's done, this PR should get merged.

All done.

louis-z (approved):

PR #655, opened 2024-04-04
Add async CRUD methods to `IProjectClient`

belcher-rok:
- Adds async CRUD methods to `IProjectClient`.
- Improves the Mock ProjectClient to better emulate actual GitLab behavior:
  - Project Create must specify Name or Path.
  - Auto-generate missing Path from Name, and vice-versa.
  - Cannot create project with duplicate Name (case-sensitive).
  - Cannot create project with duplicate Path (case-INsensitive).
  - `Project.BuildTimeout` is in seconds, not minutes.
  - GitLab [ignores](https://docs.gitlab.com/ee/api/projects.html#create-project) `ProjectCreate.DefaultBranch` unless `ProjectCreate.InitializeWithReadme` is true.
  - `Project.ToClientProject()` did not populate `NameWithNamspace`.
  - `Project.FullName` contains spaces around the slash separator.
- Added tests to verify GitLab and Mock behaviors.

### DIFF
--- a/NGitLab.Mock/Config/GitLabHelpers.cs
+++ b/NGitLab.Mock/Config/GitLabHelpers.cs
@@ -1250,10 +1250,10 @@ public static class GitLabHelpers
 
     private static void CreateProject(GitLabServer server, GitLabProject project)
     {
-        var prj = new Project(project.Name ?? project.Path ?? Guid.NewGuid().ToString("D"))
+        var projectName = project.Name ?? project.Path ?? Guid.NewGuid().ToString("D");
+        var prj = new Project(projectName, project.Path)
         {
             Id = project.Id,
-            Path = project.Path,
             Description = project.Description,
             DefaultBranch = project.DefaultBranch ?? server.DefaultBranchName,
             Visibility = project.Visibility ?? project.Parent.DefaultVisibility,
--- a/NGitLab.Mock/Config/GitLabHelpers.cs
+++ b/NGitLab.Mock/Config/GitLabHelpers.cs
@@ -320,6 +320,53 @@ public static class GitLabHelpers
     }
 
     /// <summary>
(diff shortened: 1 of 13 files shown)


belcher-rok:
@Toa741 Anything I can do to help move this along?

NGitLab.Mock/Config/GitLabHelpers.cs
```diff
@@ -1203,9 +1250,10 @@ private static void CreateGroup(GitLabServer server, GitLabGroup group)
 
     private static void CreateProject(GitLabServer server, GitLabProject project)
     {
-        var prj = new Project(project.Name ?? Guid.NewGuid().ToString("D"))
+        var prj = new Project(project.Name ?? project.Path ?? Guid.NewGuid().ToString("D"))
         {
             Id = project.Id,
+            Path = project.Path,
```
Toa741: I wonder if that would result in the intended behavior:

If the GitLabProject has a null Path:
- It will pass in the constructor of Project
- It will set Path as Slug
- It will then set it as project.Path, which will be null

It's intended?

Toa741:
> @Toa741 Anything I can do to help move this along?

Hey John :), I was on vacation, everything looks good, just had one question regarding the Project init in the mock

NGitLab.Mock/Config/GitLabHelpers.cs
belcher-rok: Good observation. I'm glad to know you are paying attention! 😉 You are correct: `config.WithProject("Test", configure: p => p.@Namespace = "Foo")` was replacing the default path with null. I have corrected this. However, it ultimately didn't matter because `Mock.ProjectCollection.Add()` auto-generates the path from the name, and vice-versa, just like GitLab does.

PR #659, opened 2024-04-11
Implements parameters for RepositoryClient.GetArchive

steve85:
Implements the `format`, `path` and `sha` options for RepositoryClient.GetArchive().

https://docs.gitlab.com/ee/api/repositories.html#get-file-archive

Allows us to retrieve the archive for a particular SHA/ref and specify the required download format.


NGitLab/IRepositoryClient.cs
```diff
@@ -21,7 +21,7 @@ public interface IRepositoryClient
 
     void GetRawBlob(string sha, Action<Stream> parser);
 
-    void GetArchive(Action<Stream> parser);
+    void GetArchive(Action<Stream> parser, string sha = null, string format = null);
```
louis-z: This is a breaking change.

Furthermore, we may eventually want to add support for a subpath specification (see [here](https://docs.gitlab.com/ee/api/repositories.html#get-file-archive)).

So I suggest the following:

1. Introduce a `FileArchiveQuery` type:
    ```cs
    public sealed class FileArchiveQuery
    {
        public Sha1 Sha { get; set; }
        public ArchiveFormat Format { get; set; }
    }
    ```
    where `ArchiveFormat` would be an `enum` of the available format options
1. Add a `GetArchive(Action<Stream> parser, FileArchiveQuery query)` overload

This way, the API will be more future-proof.

louis-z (changes requested):

NGitLab/IRepositoryClient.cs
steve85: Thanks for the detailed feedback @louis-z. I will make the requested changes and push an update for you to review.

NGitLab/Models/FileArchiveFormat.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System.Runtime.Serialization;
+
+namespace NGitLab.Models;
+
+public enum FileArchiveFormat
+{
+    [EnumMember(Value = ".bz2")]
```
steve85, line 7: I've used `EnumMemberAttribute` here to map the enum value to the required file extension. I couldn't find anywhere else in the project where you were doing something like this. 

Previously in other projects I have used `System.ComponentModel.DescriptionAttribute` for this kind of thing, but if there is a way you'd prefer me to implement this here, let me know and I'll refactor to suit.

NGitLab/Extensions/TypeExtensions.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using System.Linq;
+using System.Reflection;
+using System.Runtime.Serialization;
+
+namespace NGitLab.Extensions;
+
+internal static class TypeExtensions
+{
+    public static string GetEnumMemberAttributeValue<TEnum>(this TEnum value)
```
steve85: I also had to add this extension method to support getting the `EnumMemberAttribute` value from the enum value for the file extension translation.

NGitLab/IRepositoryClient.cs
steve85: @louis-z - I have made the requested changes but am unsure if there is a preferred way of handling the file extension translation. Please see my comments below regarding this.

NGitLab/Models/FileArchiveFormat.cs
louis-z: Using `EnumMemberAttribute` is the way to go.

NGitLab/Extensions/TypeExtensions.cs
louis-z: I was sure `Utils.AddParameter` already handled the `EnumMemberAttribute`, but I was wrong.

I made the required changes in https://github.com/ubisoft/NGitLab/pull/660.

steve85: I've rebased from main and made some further changes. 

I am unable to use `Utils.AddParameter` in this instance because in the case of the archive endpoint, the requested extension needs to be appended directly to the path when a specific format is required, i.e.:
- `/archive` - when requesting the default archive format (currently defaults to *tar.gz*)
- `/archive.ext` - when requesting a specific format where `ext` is one of the GitLab supported formats

To cater for this I've added a new method `Utils.AppendSegmentToUrl` which can be used to append a segment to the URL. I've also added a suite of unit tests for this new method. 

Please let me know if there are any issues or further changes you'd like me to make.

NGitLab.Tests/Impl/UtilsTests.cs
```diff
@@ -15,4 +15,118 @@ public void AddParameter_ConsidersEnumMemberAttribute(EventAction value, string
...
+        // Arrange
+        const string basePath = "https://gitlab.org/api/v4/stuff";
+        var url = basePath;
+        var expected = basePath;
+
+        // Act
+        var actual = NGitLab.Impl.Utils.AppendSegmentToUrl<string>(url, value: null, includeSegmentSeparator: false);
+
+        // Assert
+        Assert.That(expected, Is.EqualTo(actual));
```
louis-z: Please invert `actual` and `expected`, in conformance with the NUnit Constraint Model's API + do the same where applicable.

steve85:
@louis-z can this PR now be merged or are there further outstanding changes you'd like me to make?

louis-z:
@steve85, this PR is stale. Is there a plan to resurrect it any time soon? If not, we'll consider closing it.

steve85:
Hi @louis-z - I was originally waiting on a follow up review from you, hence the lack of activity on this. 

I have rebased the branch and made some changes to fix some minor things. Please take a look and let me know if there are any further issues that need addressing.

louis-z (changes requested):

louis-z:
> Hi @louis-z - I was originally waiting on a follow up review from you, hence the lack of activity on this.
> 
> I have rebased the branch and made some changes to fix some minor things. Please take a look and let me know if there are any further issues that need addressing.

I wrote a comment back in 2024 but forgot to submit it... 🙄 Sorry about that.

It's [this one](https://github.com/ubisoft/NGitLab/pull/659#discussion_r1565888722).

NGitLab/Models/FileArchiveFormat.cs
```diff
@@ -0,0 +1,33 @@
...
+    Bz2,
+
+    [EnumMember(Value = ".gz")]
```
louis-z: Is the `.gz` format truly supported? I don't see it mentioned in the [GitLab doc](https://docs.gitlab.com/api/repositories/#retrieve-file-archive-from-a-repository), but then again it may be outdated...

NGitLab/Impl/Utils.cs
```diff
@@ -103,11 +105,55 @@ public static string AddPageParams(string url, int? page, int? perPage)
         return url;
     }
 
+    public static string AppendSegmentToUrl<T>(string url, T value, bool includeSegmentSeparator = true)
+    {
+        if (value is null)
+            return url;
+
+        // Don't allow segments to a url which already has parameters present
```
louis-z:
```suggestion
        // Don't allow appending segments to a url which already has parameters present
```

NGitLab/Impl/Utils.cs
```diff
@@ -103,11 +105,55 @@ public static string AddPageParams(string url, int? page, int? perPage)
...
+
     private static string AddParameterInternal(string url, string parameterName, string stringValue)
     {
         var @operator = !url.Contains("?") ? "?" : "&";
         var formattedValue = WebUtility.UrlEncode(stringValue);
         var parameter = $"{@operator}{parameterName}={formattedValue}";
         return url + parameter;
     }
+
+    private static string GetEnumMemberValue<T>(string valueString)
```
louis-z: Do you really need this method? I have a feeling you could use `ToValueString<T>(T value)` instead.

NGitLab.Tests/Impl/UtilsTests.cs
```diff
@@ -16,6 +17,121 @@ public void AddParameter_ConsidersEnumMemberAttribute(EventAction value, string
         Assert.That(url, Is.EqualTo($"{basePath}?event_action={expectedQueryParamValue}"));
     }
 
+    [TestCase]
```
louis-z: Use `[Test]` here instead and for any other non-parameterized test method.

louis-z (changes requested):

NGitLab.Tests/Impl/UtilsTests.cs
steve85: Fixed.

NGitLab/Impl/Utils.cs
steve85: You're right, this method wasn't needed. I've removed it.

NGitLab/Impl/Utils.cs
steve85: Fixed.

NGitLab/Models/FileArchiveFormat.cs
steve85: It's been so long since I wrote the original code I'm not sure what I was referencing when I added `.gz`. I have removed it.

steve85:
@louis-z - I made the requested changes last week. Can you please review these changes and let me know if there is anything else required?

PR #661, opened 2024-04-23
Add support to Group hooks

### LINKED ISSUE #654: Add support to Group hooks
Currently, the library only handles project hooks, and we haven't extended support to group level hooks just yet. 

It's something we're considering, especially since it's important to account for group level hooks when retrieving a project's hooks.

Our main focus right now is on _reading_ hooks rather than _creating_ or _editing_ them.

### Additional Information
- GitLab [API Documentation](https://docs.gitlab.com/ee/api/groups.html#list-group-hooks)

PrashanthiRamesh:
- Closes #654

### DIFF
--- a/NGitLab/IGroupsClient.cs
+++ b/NGitLab/IGroupsClient.cs
@@ -188,6 +188,4 @@ public interface IGroupsClient
     /// <param name="id">The group's id.</param>
     /// <param name="cancellationToken">The cancellation token.</param>
     Task RestoreAsync(int id, CancellationToken cancellationToken = default);
-
-    IGroupHooksClient GetGroupHooks(GroupId groupId);
 }
--- a/NGitLab/IGroupsClient.cs
+++ b/NGitLab/IGroupsClient.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
+using NGitLab.Impl;
 using NGitLab.Models;
 
 namespace NGitLab;
@@ -188,4 +189,6 @@ public interface IGroupsClient
     /// <param name="id">The group's id.</param>
     /// <param name="cancellationToken">The cancellation token.</param>
     Task RestoreAsync(int id, CancellationToken cancellationToken = default);
+
+    IGroupHooksClient GetGroupHooks(GroupId groupId);
 }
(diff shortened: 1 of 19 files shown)


NGitLab/IGroupsClient.cs
```diff
@@ -188,4 +188,6 @@ public interface IGroupsClient
     /// <param name="id">The group's id.</param>
     /// <param name="cancellationToken">The cancellation token.</param>
     Task RestoreAsync(int id, CancellationToken cancellationToken = default);
+
+    IGroupHooksClient GetGroupHooks(GroupId groupId);
```
louis-z: I think this method should be declared in `IGitLabClient` instead, similar to `GetGroupBadgeClient`.

NGitLab/Models/GroupHook.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System;
+using System.Text.Json.Serialization;
+
+namespace NGitLab.Models;
+
+public class GroupHook
+{
+    [JsonPropertyName("id")]
+    public int Id;
```
louis-z: Define members as properties, not fields.

NGitLab/Models/GroupHookUpsert.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.ComponentModel.DataAnnotations;
+using System.Text.Json.Serialization;
+
+namespace NGitLab.Models;
+
+public class GroupHookUpsert
+{
+    [Required]
+    [JsonPropertyName("url")]
+    public Uri Url;
```
louis-z: Define members as properties, not fields.

louis-z (changes requested):

NGitLab/Models/GroupHookUpsert.cs
```diff
@@ -0,0 +1,39 @@
...
+namespace NGitLab.Models;
+
+public class GroupHookUpsert
+{
+    [Required]
+    [JsonPropertyName("url")]
+    public Uri Url { get; set; }
+
+    [JsonPropertyName("push_events")]
+    public bool PushEvents { get; set; }
```
louis-z: According to the [GitLab documentation](https://docs.gitlab.com/ee/api/groups.html#edit-group-hook), `push_events` is not required. I would make nullable the properties that are not required. Otherwise when updating a hook the API call could inadvertently overwrite settings that ought not to be.

louis-z (changes requested):

PR #671, opened 2024-05-06
Remove StyleCop.Analyzers

louis-z:


Toa741:
To prevent the dependabot failing on this dependency

PR #676, opened 2024-05-09
Fix project and group resolution in mocks

ap0llo:
I noticed that getting a group or project client from the mock GitLabClient by passing in a group or project path string was throwing a NullReferenceException.

This was due to the fact that each mock client that was using `FindProject()` or `FindGroup()` was passing in `id.ValueAsUriParameter()` which escapes and URL characters.
When the group or project path contained a "/", no matching entry was found.

This MR resolves this by using `ValueAsString()` instead of `ValueAsUriParameter()`.
I have also added tests that ensure a mock client can be retrieved for a group or project either by path or by the numeric id.

### DIFF
--- a/NGitLab.Mock.Tests/GitLabClientMockTest.cs
+++ b/NGitLab.Mock.Tests/GitLabClientMockTest.cs
@@ -1,0 +1,50 @@
+﻿using System;
+using System.Collections;
+using System.Linq;
+using NGitLab.Mock.Config;
+using NGitLab.Models;
+using NUnit.Framework;
+
+namespace NGitLab.Mock.Tests;
+
+
+public class GitLabClientMockTest
+{
+    public static IEnumerable TestCases
+    {
+        get
+        {
+            static TestCaseData TestCase(string name, Func<IGitLabClient, ProjectId, object> getClient) => new TestCaseData(getClient).SetArgDisplayNames(name);
+
+            // Enumerate all methods on IGitLabClient that accept a single ProjectId parameter
+
+            var methods = typeof(IGitLabClient).GetMethods()
+                .Where(method => method.IsPublic)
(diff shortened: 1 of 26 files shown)


NGitLab.Mock.Tests/GitLabClientMockTest.cs
```diff
@@ -0,0 +1,126 @@
...
+    {
+        get
+        {
+            static TestCaseData TestCase(string name, Func<IGitLabClient, GroupId, object> getClient) => new TestCaseData(getClient).SetArgDisplayNames(name);
+
+            // Create a test case for each method in IGitLabClient that has a single parameter of type GroupId
+            foreach (var method in GetMethods<GroupId>())
+            {
+                // GetGroupMergeRequest() is not implemented in the mock MergeRequestClient and will throw a NotImplementedException
+                // To avoid a test failure, exclude this mehtod from the test cases (this is the only such case)
```
Toa741:
```suggestion
                // To avoid a test failure, exclude this method from the test cases (this is the only such case)
```

NGitLab.Mock.Tests/GitLabClientMockTest.cs
```diff
@@ -0,0 +1,126 @@
...
+                // This can be removed if a mock for group merge requests is implemented
+                if (string.Equals(method.Name, nameof(IGitLabClient.GetGroupMergeRequest), StringComparison.Ordinal))
```
Toa741, line 41: To avoid to miss tests if the `MergeRequestClient` ctor with GroupId is implemented at some point, can you also add a comment in the ctor above the throw? Stating that this exlusion should be removed.

Toa741:
Thanks for your contribution :),
Indeed, it seems most of the existing tests were either using simple names or Ids.

Just a few comments, the rest looks goods

ap0llo:
Thanks for the review.

I incorporated your feedback

PR #682, opened 2024-05-17
Support gitlab from gitlab-ee:16.6.2-ee.0 to gitlab-ee:16.11.2-ee.0

fhebel:
Ciao ragazzi,

I'd like to contribute the version bump to the latest version of GitLab.

Tanti saluti

Florian

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,8 +51,10 @@ jobs:
         # Keep in sync with the version in GitLabDockerContainer.cs
         # Available tags: https://hub.docker.com/r/gitlab/gitlab-ee/tags
         gitlab: [
+          'gitlab/gitlab-ee:15.11.9-ee.0',
+          'gitlab/gitlab-ee:15.11.13-ee.0',
           'gitlab/gitlab-ee:16.6.2-ee.0',
-          'gitlab/gitlab-ee:16.11.2-ee.0',
+          'gitlab/gitlab-ee:16.11.3-ee.0',
         ]
         configuration: [ Release ]
       fail-fast: false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,6 +51,7 @@ jobs:
         # Keep in sync with the version in GitLabDockerContainer.cs
         # Available tags: https://hub.docker.com/r/gitlab/gitlab-ee/tags
         gitlab: [
+          'gitlab/gitlab-ee:16.6.2-ee.0',
           'gitlab/gitlab-ee:16.11.1-ee.0',
         ]
         configuration: [ Release ]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,8 +51,7 @@ jobs:
         # Keep in sync with the version in GitLabDockerContainer.cs
         # Available tags: https://hub.docker.com/r/gitlab/gitlab-ee/tags
         gitlab: [
-          'gitlab/gitlab-ee:15.11.9-ee.0',
-          'gitlab/gitlab-ee:15.11.13-ee.0',
+          'gitlab/gitlab-ee:16.11.1-ee.0',
         ]
         configuration: [ Release ]
       fail-fast: false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
         gitlab: [
           'gitlab/gitlab-ee:15.11.9-ee.0',
           'gitlab/gitlab-ee:15.11.13-ee.0',
-          'gitlab/gitlab-ee:16.6.2-ee.0',
+          'gitlab/gitlab-ee:16.10.6-ee.0',
           'gitlab/gitlab-ee:16.11.3-ee.0',
         ]
         configuration: [ Release ]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
         # Available tags: https://hub.docker.com/r/gitlab/gitlab-ee/tags
         gitlab: [
           'gitlab/gitlab-ee:16.6.2-ee.0',
-          'gitlab/gitlab-ee:16.11.1-ee.0',
+          'gitlab/gitlab-ee:16.11.2-ee.0',
         ]
         configuration: [ Release ]
       fail-fast: false
--- a/NGitLab.Tests/Docker/GitLabDockerContainer.cs
+++ b/NGitLab.Tests/Docker/GitLabDockerContainer.cs
@@ -28,8 +28,6 @@ public class GitLabDockerContainer
 
     // https://hub.docker.com/r/gitlab/gitlab-ee/tags/
     public const string GitLabDockerVersion = "16.11.3-ee.0"; // Keep in sync with .github/workflows/ci.yml
-    // public const string GitLabDockerVersion = "16.10.6-ee.0"; // Keep in sync with .github/workflows/ci.yml
-    // public const string GitLabDockerVersion = "15.11.13-ee.0"; // Keep in sync with .github/workflows/ci.yml
 
     private static string s_creationErrorMessage;
     private static readonly SemaphoreSlim s_setupLock = new(initialCount: 1, maxCount: 1);
--- a/NGitLab.Tests/Docker/GitLabDockerContainer.cs
+++ b/NGitLab.Tests/Docker/GitLabDockerContainer.cs
@@ -298,15 +298,19 @@ public class GitLabDockerContainer
             var url = await GetCurrentUrl(page);
             var gitLabVersionAsNuGetVersion = NuGetVersion.Parse(GitLabDockerVersion);
 
+            // Declare supported ranges
+            var isMajorVersion15 = VersionRange.Parse("[15.0,16.0)").Satisfies(gitLabVersionAsNuGetVersion);
+            var isMajorVersion16 = VersionRange.Parse("[16.0,17.0)").Satisfies(gitLabVersionAsNuGetVersion);
+
             // Login
             if (url == "/users/sign_in")
             {
-                if (VersionRange.Parse("[16.0,17.0)").Satisfies(gitLabVersionAsNuGetVersion))
+                if (isMajorVersion16)
                 {
                     await page.Locator("form[data-testid='sign-in-form'] input[name='user[login]']").FillAsync(AdminUserName);
                     await page.Locator("form[data-testid='sign-in-form'] input[name='user[password]']").FillAsync(AdminPassword);
                 }
-                else if (VersionRange.Parse("[15.0,16.0)").Satisfies(gitLabVersionAsNuGetVersion))
+                else if (isMajorVersion15)
                 {
                     await page.Locator("form#new_user input[name='user[login]']").FillAsync(AdminUserName);
                     await page.Locator("form#new_user input[name='user[password]']").FillAsync(AdminPassword);
@@ -335,13 +339,13 @@ public class GitLabDockerContainer
 
                 await page.GotoAsync(GitLabUrl + "/-/profile/personal_access_tokens");
 
-                if (VersionRange.Parse("[16.0,17.0)").Satisfies(gitLabVersionAsNuGetVersion))
+                if (isMajorVersion16)
                 {
                     await page.Locator("main[id='content-body'] button[data-testid='add-new-token-button']").ClickAsync(new LocatorClickOptions { Timeout = 5_000 });
                     formLocator = page.Locator("main[id='content-body'] form[id='js-new-access-token-form']");
                     await formLocator.Locator("input[data-testid='access-token-name-field']").FillAsync(tokenName);
                 }
-                else if (VersionRange.Parse("[15.0,16.0)").Satisfies(gitLabVersionAsNuGetVersion))
+                else if (isMajorVersion15)
                 {
                     // Try the "old" 15.x.y way
                     formLocator = page.Locator("main#content-body form");
--- a/NGitLab.Tests/Docker/GitLabDockerContainer.cs
+++ b/NGitLab.Tests/Docker/GitLabDockerContainer.cs
@@ -26,7 +26,7 @@ public class GitLabDockerContainer
     public const string ImageName = "gitlab/gitlab-ee";
 
     // https://hub.docker.com/r/gitlab/gitlab-ee/tags/
-    public const string GitLabDockerVersion = "16.11.2-ee.0"; // Keep in sync with .github/workflows/ci.yml
+    public const string GitLabDockerVersion = "16.11.3-ee.0"; // Keep in sync with .github/workflows/ci.yml
 
     private static string s_creationErrorMessage;
     private static readonly SemaphoreSlim s_setupLock = new(initialCount: 1, maxCount: 1);
--- a/NGitLab.Tests/Docker/GitLabDockerContainer.cs
+++ b/NGitLab.Tests/Docker/GitLabDockerContainer.cs
@@ -15,6 +15,7 @@ using Docker.DotNet;
 using Docker.DotNet.Models;
 using Microsoft.Playwright;
 using NGitLab.Models;
+using NuGet.Versioning;
 using NUnit.Framework;
 using Polly;
 
@@ -27,6 +28,8 @@ public class GitLabDockerContainer
 
     // https://hub.docker.com/r/gitlab/gitlab-ee/tags/
     public const string GitLabDockerVersion = "16.11.3-ee.0"; // Keep in sync with .github/workflows/ci.yml
+    // public const string GitLabDockerVersion = "16.10.6-ee.0"; // Keep in sync with .github/workflows/ci.yml
+    // public const string GitLabDockerVersion = "15.11.13-ee.0"; // Keep in sync with .github/workflows/ci.yml
 
     private static string s_creationErrorMessage;
     private static readonly SemaphoreSlim s_setupLock = new(initialCount: 1, maxCount: 1);
@@ -322,21 +325,23 @@ public class GitLabDockerContainer
 
                 var tokenName = "GitLabClientTest-" + DateTime.UtcNow.ToString("yyyyMMdd-HHmmss", CultureInfo.InvariantCulture);
 
+                var gitLabVersionAsNuGetVersion = NuGetVersion.Parse(GitLabDockerVersion);
+
                 await page.GotoAsync(GitLabUrl + "/-/profile/personal_access_tokens");
 
-                try
+                if (VersionRange.Parse("[16.0,17.0)").Satisfies(gitLabVersionAsNuGetVersion))
                 {
                     await page.Locator("main[id='content-body'] button[data-testid='add-new-token-button']").ClickAsync(new LocatorClickOptions { Timeout = 5_000 });
                     formLocator = page.Locator("main[id='content-body'] form[id='js-new-access-token-form']");
                     await formLocator.Locator("input[data-testid='access-token-name-field']").FillAsync(tokenName);
                 }
-                catch
+                else if (VersionRange.Parse("[15.0,16.0)").Satisfies(gitLabVersionAsNuGetVersion))
                 {
                     // Try the "old" 15.x.y way
-                    await page.GotoAsync(GitLabUrl + "/-/profile/personal_access_tokens");
                     formLocator = page.Locator("main#content-body form");
                     await formLocator.GetByLabel("Token name").FillAsync(tokenName);
                 }
+                else { throw new NotImplementedException($"Library not suitable for the GitLab version '{GitLabDockerVersion}'."); }
 
                 foreach (var checkbox in await formLocator.GetByRole(AriaRole.Checkbox).AllAsync())
                 {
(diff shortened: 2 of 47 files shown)


.github/workflows/ci.yml
```diff
@@ -51,8 +51,8 @@ jobs:
         # Keep in sync with the version in GitLabDockerContainer.cs
         # Available tags: https://hub.docker.com/r/gitlab/gitlab-ee/tags
         gitlab: [
-          'gitlab/gitlab-ee:15.11.9-ee.0',
-          'gitlab/gitlab-ee:15.11.13-ee.0',
+          'gitlab/gitlab-ee:16.6.2-ee.0',
```
louis-z: Ideally we would also keep on supporting 15.11.*.

NGitLab.Tests/MergeRequest/MergeRequestClientTests.cs
```diff
@@ -50,10 +50,7 @@ public async Task Test_merge_request_api()
         var commits = mergeRequestClient.Commits(mergeRequest.Iid).All;
         Assert.That(commits.Any(), Is.True, "Can return the commits");
 
-        if (context.IsGitLabVersionInRange(VersionRange.Parse("[15.6,)"), out _))
-            Assert.That(mergeRequest.DetailedMergeStatus.EnumValue, Is.Not.Null);
-        else
-            Assert.That(mergeRequest.DetailedMergeStatus.EnumValue, Is.Null);
+        Assert.That(mergeRequest.DetailedMergeStatus.EnumValue, Is.Null);
```
louis-z: The previous code actually told us that, starting at version 15.6, a change in behavior was introduced. I would tend to keep it for now. At least, while still officially supporting versions 15.x.y...

louis-z:
Thanks @fhebel, your contribution is much appreciated!

NGitLab/Impl/UserClient.cs
```diff
@@ -110,7 +110,7 @@ public void Deactivate(int userId)
     public GitLabCollectionResponse<LastActivityDate> GetLastActivityDatesAsync(DateTimeOffset? from = null)
     {
         var url = "/user/activities";
-        if (from is not null)
+        if (from != null)
```
louis-z: Please keep the pattern matching condition, as it's more readable and _(though unnecessary here)_ `the compiler guarantees that no user-overloaded == or != operator is invoked` according to [Microsoft's documentation](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/is).

louis-z (changes requested):

fhebel:
Hi @louis-z 

thanks a lot for your fast feedback.

I’ll have to think how to still support the 15.11.* because in meantime some deprecated functionality has been removed in GitLab (also in the interfaces). Maybe you have advice how to address such issues from your experience in the past?

Mille grazie in anticipo

Tanti saluti

Florian

NGitLab.Tests/Docker/GitLabDockerContainer.cs
```diff
@@ -320,10 +320,11 @@ async Task GenerateAdminToken(GitLabCredential credentials)
 
                 await page.GotoAsync(GitLabUrl + "/-/profile/personal_access_tokens");
 
-                var formLocator = page.Locator("main#content-body form");
+                await page.Locator("main[id='content-body'] button[data-testid='add-new-token-button']").ClickAsync();
+                var formLocator = page.Locator("main[id='content-body'] form[id='js-new-access-token-form']");
 
                 var tokenName = "GitLabClientTest-" + DateTime.UtcNow.ToString("yyyyMMdd-HHmmss", CultureInfo.InvariantCulture);
-                await formLocator.GetByLabel("Token name").FillAsync(tokenName);
+                await formLocator.Locator("input[data-testid='access-token-name-field']").FillAsync(tokenName);
```
louis-z: To initialize the GitLab server for versions 15 and 16, the following code seems to work _(though probably not optimal)_:

```suggestion
                var tokenName = "GitLabClientTest-" + DateTime.UtcNow.ToString("yyyyMMdd-HHmmss", CultureInfo.InvariantCulture);

                await page.GotoAsync(GitLabUrl + "/-/profile/personal_access_tokens");

                ILocator formLocator;
                try
                {
                    await page.Locator("main[id='content-body'] button[data-testid='add-new-token-button']").ClickAsync(new LocatorClickOptions { Timeout = 5_000 });
                    formLocator = page.Locator("main[id='content-body'] form[id='js-new-access-token-form']");
                    await formLocator.Locator("input[data-testid='access-token-name-field']").FillAsync(tokenName);
                }
                catch
                {
                    // Try the "old" 15.x.y way
                    await page.GotoAsync(GitLabUrl + "/-/profile/personal_access_tokens");
                    formLocator = page.Locator("main#content-body form");
                    await formLocator.GetByLabel("Token name").FillAsync(tokenName);
                }
```

fhebel: Proposal committed
[fcdaa4f](https://github.com/ubisoft/NGitLab/pull/682/commits/fcdaa4fbb86fd5be718c7f9121d37a2963f2c5b9)

NGitLab/Impl/UserClient.cs
fhebel: [0cf0975](https://github.com/ubisoft/NGitLab/pull/682/commits/0cf09752113b6c122425856f515b152a978a7e83)

.github/workflows/ci.yml
fhebel: [70b831d](https://github.com/ubisoft/NGitLab/pull/682/commits/70b831d39b2eb5dbbfe74cbfa903397ec16b85e5)

NGitLab/Models/ReleaseLink.cs
```diff
@@ -29,9 +29,6 @@ public class ReleaseLink
     [JsonPropertyName("direct_asset_url")]
     public string DirectAssetUrl { get; set; }
 
-    [JsonPropertyName("external")]
```
Toa741, line 32: As we still support the v15 that still have it, instead of removing it or marking it obsolete, we usually hide those with `[EditorBrowsable(EditorBrowsableState.Never)]`. It makes them non browsable but still usable if it was already used.

NGitLab/Models/DetailedMergeStatus.cs
```diff
@@ -33,4 +33,7 @@ public enum DetailedMergeStatus
     NotOpen,
     [EnumMember(Value = "policies_denied")]
     PoliciesDenied,
+    // Undocumented member
+    [EnumMember(Value = "preparing")]
```
Toa741, line 37: Just to know, where did you find the value?

NGitLab/IEnvironmentClient.cs
```diff
@@ -22,12 +22,13 @@ public interface IEnvironmentClient
...
     /// Updates an existing environment's name and/or external_url.
+    /// Renaming an environment by using the API was deprecated in GitLab 15.9.
+    /// Renaming an environment with the API removed in GitLab 16.0.
     /// </summary>
     /// <param name="environmentId">The ID of the environment</param>
-    /// <param name="name">The new name of the environment</param>
     /// <param name="externalUrl">The new external url</param>
     /// <returns>The updated environment</returns>
-    EnvironmentInfo Edit(int environmentId, string name, string externalUrl);
+    EnvironmentInfo Edit(int environmentId, string externalUrl);
```
Toa741, line 44: Breaking change.

Can you let the old method, with `[EditorBrowsable(EditorBrowsableState.Never)]`. Then you can let the new one without the name param?

And in the implementation, you can just redirect the new one to the old one with a null name.

NGitLab.Tests/UsersTests.cs
```diff
@@ -229,11 +229,19 @@ public async Task GetLastActivityDatesFromTheFuture()
...
-        var exception = Assert.Throws<GitLabException>(() => context.Client.Users.GetLastActivityDatesAsync().ToArray());
-        Assert.That(exception.StatusCode, Is.EqualTo(HttpStatusCode.Forbidden));
+        var privateProfileUsers = context.AdminClient.Users.All.Where(u => u.PrivateProfile).Select(ppu => ppu.Username).ToList();
+        Assert.That(privateProfileUsers, Is.Not.Empty);
+
+        var lastActivityDates = context.Client.Users.GetLastActivityDatesAsync().ToList();
+        Assert.That(lastActivityDates, Is.Not.Empty);
+
+#pragma warning disable MA0002
+        var lastActivityDatesOfPrivateProfileUsers = lastActivityDates.Where(lad => privateProfileUsers.Contains(lad.Username)).ToList();
```
Toa741: Isn't it just adding StringComparison.Ordinal or StringComparer.Ordinal to the contains method to avoid the warning?

NGitLab.Tests/Docker/GitLabDockerContainer.cs
```diff
@@ -316,14 +316,27 @@ async Task GenerateAdminToken(GitLabCredential credentials)
...
 
-                var tokenName = "GitLabClientTest-" + DateTime.UtcNow.ToString("yyyyMMdd-HHmmss", CultureInfo.InvariantCulture);
-                await formLocator.GetByLabel("Token name").FillAsync(tokenName);
+                try
+                {
+                    await page.Locator("main[id='content-body'] button[data-testid='add-new-token-button']").ClickAsync(new LocatorClickOptions { Timeout = 5_000 });
+                    formLocator = page.Locator("main[id='content-body'] form[id='js-new-access-token-form']");
+                    await formLocator.Locator("input[data-testid='access-token-name-field']").FillAsync(tokenName);
+                }
+                catch
```
Toa741: Could the exception raised be specified for clarity (and to avoid catching others)? 

Is the GotoAsync in the catch necessary too? As you would already be on the correct page.

NGitLab/Models/ReleaseLink.cs
fhebel: Hi @Toa741 ,

thanks a lot for that advice!

NGitLab/Models/DetailedMergeStatus.cs
fhebel: The test case "Test_merge_request_api()" failed for version 16, because the enum value was null, but the string was "preparing":
Assert.That(mergeRequest.DetailedMergeStatus.EnumValue, Is.Not.Null);

PS: Someone else reported that behaviour as well: https://forum.gitlab.com/t/rest-api-get-single-mr-returns-undocumented-detailed-merge-status/98122

NGitLab/Models/ReleaseLink.cs
fhebel: Done

NGitLab/IEnvironmentClient.cs
fhebel: Done

NGitLab.Tests/UsersTests.cs
fhebel: You're completely right. I changed it ...

NGitLab.Tests/Docker/GitLabDockerContainer.cs
fhebel: You're completely right. I just had taken @louis-z's proposal. In the new implementation I'm just checking the version number, which we're running against ...

NGitLab.Tests/Docker/GitLabDockerContainer.cs
```diff
@@ -295,19 +298,29 @@ async Task GenerateAdminToken(GitLabCredential credentials)
...
 
             var url = await GetCurrentUrl(page);
+            var gitLabVersionAsNuGetVersion = NuGetVersion.Parse(GitLabDockerVersion);
 
             // Login
             if (url == "/users/sign_in")
             {
-                await page.Locator("form#new_user input[name='user[login]']").FillAsync(AdminUserName);
-                await page.Locator("form#new_user input[name='user[password]']").FillAsync(AdminPassword);
+                if (VersionRange.Parse("[16.0,17.0)").Satisfies(gitLabVersionAsNuGetVersion))
```
Toa741: Could you make the range to be version 16 and upper (not limiting between 16 and 17), cause it would make a potential version 18 to throw (and ideally they don't change their html markup again on future major versions ^^).

Should maybe be `[16.0,)` ?

Also, as the version ranges are repeated, could we have something like:
```csharp
// Declare supported ranges
var isMajorVersion15 = VersionRange.Parse("[15.0,16.0)").Satisfies(gitLabVersionAsNuGetVersion);
var isMajorVersionAtLeast16 = VersionRange.Parse("[16.0,)").Satisfies(gitLabVersionAsNuGetVersion);

// Then, instead of having to throw at 2 places, it can be checked early
if  (!isMajorVersion15  && !isMajorVersionAtLeast16)
{
    throw new NotImplementedException($"Unable to generate an admin token: selected GitLabDockerVersion '{GitLabDockerVersion}' doesn't match any supported range.")
}

// Then variables can be reused for the conditions
if (isMajorVersionAtLeast16)
{
  // playwright code for 16+
}
else
{
  // playwright code for 15
}
```

NGitLab.Tests/Docker/GitLabDockerContainer.cs
```diff
@@ -26,7 +27,9 @@ public class GitLabDockerContainer
     public const string ImageName = "gitlab/gitlab-ee";
 
     // https://hub.docker.com/r/gitlab/gitlab-ee/tags/
-    public const string GitLabDockerVersion = "15.11.9-ee.0"; // Keep in sync with .github/workflows/ci.yml
+    public const string GitLabDockerVersion = "16.11.3-ee.0"; // Keep in sync with .github/workflows/ci.yml
+    // public const string GitLabDockerVersion = "16.10.6-ee.0"; // Keep in sync with .github/workflows/ci.yml
+    // public const string GitLabDockerVersion = "15.11.13-ee.0"; // Keep in sync with .github/workflows/ci.yml
```
Toa741:
```suggestion
    public const string GitLabDockerVersion = "16.11.3-ee.0"; // Keep in sync with .github/workflows/ci.yml
```

I don't think it's really needed to let the examples here

NGitLab.Tests/EnvironmentsTests.cs
```diff
@@ -84,15 +85,19 @@ public async Task Edit()
         // Validate newEnvNameToEdit is present
         Assert.That(envClient.All.FirstOrDefault(e => string.Equals(e.Name, newEnvNameToEdit, StringComparison.Ordinal)), Is.Not.Null);
 
+        // Handling for old version 15.x
+        var isVersion15 = context.IsGitLabVersionInRange(VersionRange.Parse("[15.0,16.0)"), out _);
+
         // Edit and check return value
         env = envClient.Edit(initialEnvId, newEnvNameUpdated, newEnvNameExternalUrlUpdated);
-        Assert.That(env.Name, Is.EqualTo(newEnvNameUpdated).IgnoreCase);
+        if (isVersion15) { Assert.That(env.Name, Is.EqualTo(newEnvNameUpdated).IgnoreCase); }
```
Toa741:
```suggestion

        if (isVersion15)
        { 
            Assert.That(env.Name, Is.EqualTo(newEnvNameUpdated).IgnoreCase);
        }

```

Not a fan on one one line conditions, can you indent the others too?

NGitLab.Tests/EnvironmentsTests.cs
```diff
@@ -84,15 +85,19 @@ public async Task Edit()
         // Validate newEnvNameToEdit is present
         Assert.That(envClient.All.FirstOrDefault(e => string.Equals(e.Name, newEnvNameToEdit, StringComparison.Ordinal)), Is.Not.Null);
 
+        // Handling for old version 15.x
+        var isVersion15 = context.IsGitLabVersionInRange(VersionRange.Parse("[15.0,16.0)"), out _);
```
Toa741: As it's repeated in multiple places in tests, how about a little method like the existing `IsGitLabVersionInRange`:

```csharp
// In the GitLabTestContext
public bool IsGitLabMajorVersion(int major)
{
    var currentVersion = Client.Version.Get();
    return NuGetVersion.TryParse(gitLabVersion, out var nuGetVersion) && nuGetVersion.Major == major;
}

// In tests, that would look like
if (context.IsGitLabMajorVersion(15)) 
{ 
    // Additional asserts
}
```

NGitLab.Tests/Docker/GitLabDockerContainer.cs
fhebel: Hi @Toa741,

I agree with you. Unfortunately for version 17 there needs to be yet another implementation - I already tested the current code with 17.0.1 and it didn't work.

How do you suggest to proceed?

NGitLab.Tests/Docker/GitLabDockerContainer.cs
fhebel: I apologize. I commited my test code by mistake ...

NGitLab.Tests/EnvironmentsTests.cs
fhebel: sure

NGitLab.Tests/Docker/GitLabDockerContainer.cs
fhebel: Removed

NGitLab.Tests/EnvironmentsTests.cs
fhebel: Done

NGitLab.Tests/EnvironmentsTests.cs
fhebel: Done

NGitLab.Tests/Docker/GitLabDockerContainer.cs
Toa741: Oh, then in this case we can keep the range you had ([16.0,17.0))

fhebel:
Could you please give me some advice how to fix the github actions.

```csharp
          'gitlab/gitlab-ee:15.11.9-ee.0',
          'gitlab/gitlab-ee:15.11.13-ee.0',
          'gitlab/gitlab-ee:16.10.6-ee.0',
          'gitlab/gitlab-ee:16.11.3-ee.0',
```
On the first attempt only one test of version 16 passes, all the others fail. Then I restart the other of version 16 - it passes. The action regarding version 15 time out although they're running fine on my local machine.

Any advice?

NGitLab.Tests/Docker/GitLabDockerContainer.cs
fhebel: Extraction of supported ranges: Done

NGitLab.Tests/EnvironmentsTests.cs
```diff
@@ -86,13 +86,18 @@ public async Task Edit()
 
         // Edit and check return value
         env = envClient.Edit(initialEnvId, newEnvNameUpdated, newEnvNameExternalUrlUpdated);
-        Assert.That(env.Name, Is.EqualTo(newEnvNameUpdated).IgnoreCase);
+
+        if (context.IsGitLabMajorVersion(15))        {
+            Assert.That(env.Name, Is.EqualTo(newEnvNameUpdated).IgnoreCase);
+        }
```
Toa741:
```suggestion
        if (context.IsGitLabMajorVersion(15))
        {
            Assert.That(env.Name, Is.EqualTo(newEnvNameUpdated).IgnoreCase);
        }
```

Toa741:
> Could you please give me some advice how to fix the github actions.
> 
> ```cs
>           'gitlab/gitlab-ee:15.11.9-ee.0',
>           'gitlab/gitlab-ee:15.11.13-ee.0',
>           'gitlab/gitlab-ee:16.10.6-ee.0',
>           'gitlab/gitlab-ee:16.11.3-ee.0',
> ```
> 
> On the first attempt only one test of version 16 passes, all the others fail. Then I restart the other of version 16 - it passes. The action regarding version 15 time out although they're running fine on my local machine.
> 
> Any advice?

Let's rerun them, I'll also try to check locally on my side

fhebel:
> Let's rerun them, I'll also try to check locally on my side

I re-run the tests on my local machine for the 15.x GitLab versions and there was no problem.

Can you confirm the same for your local tests?

In the meantime I'll see how we can debug the github actions ...

Toa741:
> > Let's rerun them, I'll also try to check locally on my side
> 
> I re-run the tests on my local machine for the 15.x GitLab versions and there was no problem.
> 
> Can you confirm the same for your local tests?
> 
> In the meantime I'll see how we can debug the github actions ...

Seems fine locally as well on my side, so that's probably something related to when it executes on the CI. Will try to look at that as well

Toa741:
I had a doubt about having `GitLabDockerVersion` that is a const. I tested from your branch how it would behave if I lower it the the lowest supported version (https://github.com/ubisoft/NGitLab/pull/694/commits/aa8edf50b5c3b069b2dd5f3c2deb1bc5a0ea8788) and strangely enough, the tests on 15.x works now, but not the 16.x ones.

So, probably not your code, but maybe the logic around `GitLabDockerVersion` that is not good (and the problem was hidden maybe of having to support the same major before).

Toa741:
Could likely be the use of `GitLabDockerVersion` when checking the version in `GenerateAdminToken` 😄

Otohits-webmaster:
https://github.com/ubisoft/NGitLab/pull/695

Toa741:
https://github.com/ubisoft/NGitLab/pull/695

fhebel:
@Toa741 Thanks a lot for your help!

Toa741:
@fhebel, your welcome :).

Didn't had the time to write again on Friday, it make the GitLabDockerContainer class a bit clearer and should support both v15 and 16 (first time all the test are passing).

Just have one final suggestion: we should keep the lowest version on `LocalGitLabDockerVersion`

fhebel:
Once this PR is done I'd like to contribute a new VariableClient in order to support scoped variables (Create, Update, Delete) ...

louis-z (approved):

PR #695, opened 2024-06-07
Resolve GitLab version in GitLabDockerContainer

Toa741:
- Add comments to clarify `GitLabDockerVersion` property
- Retrieve GitLab version once logged in on the instance
- Store its value in `ResolvedGitLabVersion` that could be used afterwards
- Supports both v15/v16 login process as well as on token retrieval

image

### DIFF
--- a/NGitLab.Tests/Docker/GitLabDockerContainer.cs
+++ b/NGitLab.Tests/Docker/GitLabDockerContainer.cs
@@ -25,8 +25,20 @@ public class GitLabDockerContainer
     public const string ContainerName = "NGitLabClientTests";
     public const string ImageName = "gitlab/gitlab-ee";
 
-    // https://hub.docker.com/r/gitlab/gitlab-ee/tags/
-    public const string GitLabDockerVersion = "15.11.9-ee.0"; // Keep in sync with .github/workflows/ci.yml
+    /// <summary>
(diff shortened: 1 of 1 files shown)


NGitLab.Tests/Docker/GitLabDockerContainer.cs
```diff
@@ -264,8 +283,10 @@ private async Task SpawnDockerContainerAsync()
         TestContext.Progress.WriteLine("GitLab Docker container is ready");
     }
 
-    private async Task GenerateCredentialsAsync()
+    private async Task GenerateCredentialsAsync(IBrowserContext context)
```
louis-z: We have way too many context types... Do you mind renaming the parameter to `browserContext`?

NGitLab.Tests/Docker/GitLabDockerContainer.cs
```diff
@@ -451,4 +443,99 @@ private static string GetCredentialsFilePath()
     {
         return Path.Combine(Path.GetTempPath(), "ngitlab", "credentials.json");
     }
+
+    private async Task VerifyCiGitLabInstance()
```
louis-z: I would probably rename this method to `WaitForCiGitLabInstance`

PR #697, opened 2024-06-11
Add variables with scope

fhebel:
At the moment it's only supported to read ci/cd-variables with scope, now they can be created, updated and deleted ...


NGitLab/Models/Variable.cs
```diff
@@ -19,6 +27,17 @@ public class Variable
     [JsonPropertyName("masked")]
     public bool Masked { get; set; }
 
+    [JsonPropertyName("raw")]
+    public bool Raw { get; set; }
+
+    /// <summary>
+    /// The environment scope of a variable
+    /// </summary>
+    /// <returns>The environment scope of a variable</returns>
```
Toa741:
```suggestion
```

NGitLab/Models/Variable.cs
```diff
@@ -19,6 +27,17 @@ public class Variable
...
+    /// <remarks>
+    /// Create and Update of project variable: All tiers: Free, Premium, Ultimate <br/>
```
Toa741:
```suggestion
    /// Create and Update of project variable: All tiers (Free, Premium, Ultimate).<br/>
```

NGitLab/Models/Variable.cs
```diff
@@ -19,6 +27,17 @@ public class Variable
...
+    /// Create and Update of group variable: Premium and Ultimate only.
+    /// </remarks>
     [JsonPropertyName("environment_scope")]
     public string Scope { get; set; }
```
Toa741:
```suggestion
    public string EnvironmentScope { get; set; }
```

I think we could let the full name, as it's not the scope of the variable in itself, but linked to environments

NGitLab/Impl/VariableClient.cs
```diff
@@ -18,9 +20,20 @@ protected VariableClient(API api, string urlPrefix)
...
-    public Variable Update(string key, VariableUpdate model) => _api.Put().With(model).To<Variable>(_urlPrefix + "/variables/" + key);
+    public Variable Create(Variable model) => _api.Post().With(model).To<Variable>(_urlPrefix + "/variables");
 
-    public void Delete(string key) => _api.Delete().Execute(_urlPrefix + "/variables/" + key);
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public Variable Update(string key, VariableUpdate model) => _api.Put().With(model).To<Variable>($"{_urlPrefix}/variables/{key}");
+
+    public Variable Update(string key, Variable model) => Update(key, model, null);
+
+    public Variable Update(string key, Variable model, string environmentScope) => _api.Put().With(model).To<Variable>(string.IsNullOrWhiteSpace(environmentScope) ? $"{_urlPrefix}/variables/{key}" : $"{_urlPrefix}/variables/{key}?filter[environment_scope]={Uri.EscapeDataString(environmentScope!)}");
```
Toa741: As the environment scope is already a property in the Variable class, I would just take the property directly to make the condition, instead of having another param in the methods.

It might change a bit the tests, but it would feels probably more natural when using the client ^^

NGitLab/Impl/GroupVariableClient.cs
```diff
@@ -1,11 +1,12 @@
...
+using NGitLab.Models;
 
 namespace NGitLab.Impl;
 
 internal sealed class GroupVariableClient : VariableClient, IGroupVariableClient
 {
-    public GroupVariableClient(API api, GroupId groupId)
-        : base(api, $"{Group.Url}/{groupId.ValueAsUriParameter()}")
+    public GroupVariableClient(API api, GroupId groupId, string environmentScope = null)
+        : base(api, string.IsNullOrWhiteSpace(environmentScope) ? $"{Group.Url}/{groupId.ValueAsUriParameter()}" : $"{Group.Url}/{groupId.ValueAsUriParameter()}?filter[environment_scope]={Uri.EscapeDataString(environmentScope!)}")
```
Toa741: I would avoid logic in the constructor.

Can the presence of the querystring be determined in the methods directly?

NGitLab.Tests/ProjectVariableClientTests.cs
```diff
@@ -52,4 +55,99 @@ public async Task Test_project_variables()
...
+            Value = "My value",
+            Description = "Some important variable",
+            Protected = true,
+            Type = VariableType.Variable,
+            Masked = false,
+            Raw = false,
+            Scope = "test/*",
+        };
+
+        Assert.That(variable.IsMatchForEnvironment("test/"), Is.True);
```
Toa741: Seems that IsMatchForEnvironment is only used in this test.

Here you are more testing the logic of the extension method rather than the logic of GitLab.

I would favorise a test like you did above to validate the scopes

NGitLab/Models/Variable.cs
fhebel: Resolved with bf58a1c80cc5c00ba58a3519337f04dcee412149

NGitLab/Impl/VariableClient.cs
fhebel: The idea behind this signature was that this way you can actually change the environment scope of an existing variable. If I rely solely on the property EnvironmentScope I would need to delete the existing and create a new Variable.

@Toa741 What are your thoughts given that additional piece of information?

NGitLab.Tests/ProjectVariableClientTests.cs
fhebel: I added another test as showcase of the functionality c168edc26e0c5ecdad387a067a98d81b5abc92f1

@Toa741 Would you please allow me to let the other testcase in as well?

NGitLab/Impl/VariableClient.cs
Toa741: Ok, I see the reason you did it as such. I'll check again if there would be another way or if we go with that

NGitLab.Tests/ProjectVariableClientTests.cs
Toa741: What I would do:
- I would keep the test you added (`Test_matching_scoped_variables`)
- I would put a different value on the 2 different variables (but keep the same key)
- Instead of checking on our side the environment
  - We can have a first check that, when no environment is specified, it should return a 409 (conflict)
  - When specifying an environment, we can check it's the correct value of it that is returned

For me, checking the resolution of the environment is out of scope (it's the inner responsibility of GitLab, not on the client to do so, even in the tests).
So would simply keep  the modified `Test_matching_scoped_variables`, then remove completely the `Test_standalone_matching_scoped_variables` one (and the need of IsMatchForEnvironment)

NGitLab/Impl/VariableClient.cs
Toa741: I checked a bit the API:
- We need the variable Key and potentially the scope to construct the URL in any case, with its corresponding variable values
- As such, I would maybe specify the environment scope as the second param (to keep the URL item along)?
- For the Variable object to pass:
  - If we follow what you did in the test (`new Variable { Value = "My value edited", Protected = false, EnvironmentScope = newScope}`), for me it would make sense maybe to pass a separate object (like an Updated VariableUpdate class), as such we make sure that we can't just pass an existing variable that we modified directly. And it will contains only the modified properties (others will be let as null). It would behave like a 'patch'.
  - Or, we can let the possiblity to pass Variable class directly in it, though, with the key/scope to pass as well as parameters, it might be messy to understand.
- Also, I would make the param names clearer:
  - key => currentVariableKey
  - environmentScope => currentVariableEnvironmentScope
  - variable => updatedVariable (if it's a Variable class) or variableUpdate (if it's a VariableUpdate class)

@louis-z, I wonder about your input on this? The VariableUpdate sounds like there would be less misunderstanding for me.

NGitLab/Extensions/VariableExtensions.cs
```diff
@@ -0,0 +1,21 @@
+using System.Text.RegularExpressions;
+using NGitLab.Models;
+
+#pragma warning disable MA0009 // Add regex evaluation timeout
+
+namespace NGitLab.Extensions;
+
+public static class VariableExtensions
+{
+    public static bool IsMatchForEnvironment(this Variable variable, string environment)
```
louis-z: As @Toa741 pointed out, it's not NGitLab's responsability to validate environment scope. I would remove this file.

NGitLab/Impl/VariableClient.cs
louis-z: I think we should keep using `VariableCreate` and `VariableUpdate`, as this will allow us to clearly define which properties can/must be set for creation and updating. Yes, this means adding the missing properties there too, not just in the `Variable` class, but how often will this change afterwards? Not often, I think.

fhebel: Hi @louis-z,

the discussion here was a totally different one. @Toa741 suggested to leave the environmentScope (string) out of the method signature. I argued that this way you wouldn't be able to just change the environment scope of an existing variable, instead you would have to delete the old variable and create a new one.

So keeping `VariableCreate` and `VariableUpdate` makes no difference given that all properties are used anyway ...

Toa741: - Keeping the key and the environmentScope in the signature make sense, as they are required (and used in the URL)
- The change I would make it to put the  environmentScope as the second argument to keep the key and the scope close as they are part of the "query", not the update in itself

```csharp
Variable Update(string key, VariableUpdate model);
Variable Update(string key, string environmentScope, VariableUpdate model);
```

For the `VariableCreate/Update`, it could make a difference in term of visibility and to avoid missuses.
- It prevents using an existing `Variable` directly (less confusion for the update case with the current key/scope to use)
- It enforce the fact that the returned value should be used instead of a modified version of the `Variable` object
- It avoids using partial `Variable` declaration as in the test (that's what made me questioned about keeping  `VariableCreate/Update`)
- Not all properties seem used actually between a create and update (at least on v15 while playing with the API directly, I can't update the key. Didn't test on v16, but could be quickly done by writing a test :))

NGitLab/Impl/GroupVariableClient.cs
fhebel: You're totally right. I reverted the change ...

NGitLab/Impl/VariableClient.cs
fhebel: I believe this issue has been solved with b9d6ac32dd64da4912d529254e99cc2a8a7b1146

NGitLab/IProjectBadgeClient.cs
```diff
@@ -26,24 +26,32 @@ public interface IProjectVariableClient
 {
     IEnumerable<Variable> All { get; }
 
-    Variable this[string key] { get; }
+    Variable this[string key, string environmentScope = null] { get; }
```
Toa741:
```suggestion
    Variable this[string key] { get; }
    
    Variable this[string key, string environmentScope] { get; }
```

I would separated instead of using optional param, as adding it (even if optional) is a change of the signature, thus, a breaking change.

NGitLab/Impl/VariableClient.cs
```diff
@@ -8,6 +9,8 @@ internal abstract class VariableClient
     private readonly string _urlPrefix;
     private readonly API _api;
 
+    private string EnvironmentScopeFilter(string environmentScope = null) => !string.IsNullOrWhiteSpace(environmentScope) ? $"?filter[environment_scope]={Uri.EscapeDataString(environmentScope!)}" : string.Empty;
```
Toa741:
```suggestion
    private string EnvironmentScopeFilter(string environmentScope = null) => !string.IsNullOrWhiteSpace(environmentScope) ? $"?filter[environment_scope]={Uri.EscapeDataString(environmentScope)}" : string.Empty;
```

Normally the compiler should be smart enough to not report it as nullable here (due to the use of IsNullOrWhiteSpace). Was there a warning on it?

NGitLab/Models/Variable.cs
```diff
@@ -19,6 +27,16 @@ public class Variable
...
+    public bool Raw { get; set; }
+
+    /// <summary>
+    /// The environment scope of a variable
+    /// </summary>
+    /// <remarks>
+    /// Create and Update of project variable: All tiers (Free, Premium, Ultimate).<br/>
+    /// Create and Update of group variable: Premium and Ultimate only.
+    /// </remarks>
     [JsonPropertyName("environment_scope")]
```
Toa741, line 48: I may have missed this one, renaming = breaking change, though, as EnvironmentScope makes more sense on both this class and the others (Create/Update), what we can do is:

```csharp
[EditorBrowsable(EditorBrowsableState.Never)]
public string Scope
{
    get => EnvironmentScope;
    set => EnvironmentScope = value;
}

/// <summary>
/// The environment scope of a variable
/// </summary>
/// <remarks>
/// Create and Update of project variable: All tiers (Free, Premium, Ultimate).<br/>
/// Create and Update of group variable: Premium and Ultimate only.
/// </remarks>
[JsonPropertyName("environment_scope")]
public string EnvironmentScope { get; set; }
```

NGitLab/IProjectBadgeClient.cs
```diff
@@ -26,24 +26,32 @@ public interface IProjectVariableClient
...
+
+    void Delete(string key, string environmentScope);
 }
 
 public interface IGroupVariableClient
 {
     IEnumerable<Variable> All { get; }
 
-    Variable this[string key] { get; }
+    Variable this[string key, string environmentScope = null] { get; }
```
Toa741:
```suggestion
```suggestion
    Variable this[string key] { get; }
    
    Variable this[string key, string environmentScope] { get; }
```
```

File: NGitLab/Models/VariableCreate.cs [hunk 12]
```diff
@@ -1,15 +1,44 @@
...
+using System.Text.Json.Serialization;
 
 namespace NGitLab.Models;
 
+[EditorBrowsable(EditorBrowsableState.Never)]
 public class VariableCreate
 {
     [JsonPropertyName("key")]
-    public string Key;
+    public string Key { get; set; }
```
Toa741: We should keep fields (breaking change if upgraded to properties)

NGitLab/Models/VariableUpdate.cs
```diff
@@ -1,12 +1,40 @@
...
+﻿using System.ComponentModel;
+using System.Text.Json.Serialization;
 
 namespace NGitLab.Models;
 
 public class VariableUpdate
 {
     [JsonPropertyName("value")]
-    public string Value;
+    public string Value { get; set; }
```
Toa741: Should keep fields

Toa741 (commented):
For the points stating that it's a breaking change: at the moment we try to avoid them.

We have an internal list of the breaking changes that would be neat to add in a further major version. Fields to properties and cleaning old or obsolete props and methods as well.

NGitLab/Impl/VariableClient.cs
fhebel: @Toa741 You're completely right.

NGitLab/Models/Variable.cs
fhebel: Addressed by [143c88e](https://github.com/ubisoft/NGitLab/pull/697/commits/143c88eccdffc05dc0a5828776c0f572c460fd3c)

NGitLab/IProjectBadgeClient.cs
fhebel: Addressed by 824f7801eb4603bed79d45ba78a85e11d4212514

NGitLab/Models/VariableCreate.cs
fhebel: Addressed by b153ebe

NGitLab/Models/VariableUpdate.cs
fhebel: Addressed by b153ebe

fhebel:
@Toa741 Thanks again for the fabulous collaboration!

PR #703, opened 2024-06-21
Support major GitLab versions greater than 16

fhebel:
> [Toa741](https://github.com/Toa741) [3 weeks ago](https://github.com/ubisoft/NGitLab/pull/682#discussion_r1621243418)
Could you make the range to be version 16 and upper (not limiting between 16 and 17), cause it would make a potential version 18 to throw (and ideally they don't change their html markup again on future major versions ^^).

> Should maybe be [16.0,) ?

@Toa741 I apologize, you were right. It seems I tested when there were still other problems ...


Toa741:
No problem, we will surely need to adapt that depending on the next versions ^^.

Can you rebase your branch? I'll launch the test after

PR #715, opened 2024-07-09
Add project ancestor groups logic

BallyB:
Adding the logic to retrieve ancestor group(s) of a specific project. See https://docs.gitlab.com/ee/api/projects.html#list-a-projects-groups

### DIFF
--- a/NGitLab.Tests/ProjectsTests.cs
+++ b/NGitLab.Tests/ProjectsTests.cs
@@ -766,7 +766,6 @@ public class ProjectsTests
 
         var groups = projectClient.GetGroupsAsync(project.Id, new ProjectGroupsQuery()).ToArray();
 
-        Assert.That(groups.Select(g => g.Id), Contains.Item(group.Id));
-        Assert.That(groups.Select(g => g.Id), Contains.Item(subgroup.Id));
+        Assert.That(groups.Select(g => g.Id), Is.EquivalentTo(new int[] { group.Id, subgroup.Id }));
     }
 }
--- a/NGitLab.Tests/ProjectsTests.cs
+++ b/NGitLab.Tests/ProjectsTests.cs
@@ -746,7 +746,7 @@ public class ProjectsTests
         var project = context.CreateProject(group.Id);
         context.CreateGroup();
 
-        var groups = projectClient.GetGroupsAsync(group.Id, new ProjectGroupsQuery
+        var groups = projectClient.GetGroupsAsync(project.Id, new ProjectGroupsQuery
         {
             Search = group.Name,
         }).ToArray();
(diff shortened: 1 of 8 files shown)


NGitLab.Tests/ProjectsTests.cs
```diff
@@ -735,4 +735,38 @@ public async Task CreateProjectWithSquashOption(SquashOption? inputSquashOption)
...
+        using var context = await GitLabTestContext.CreateAsync();
+        var projectClient = context.Client.Projects;
+        var group = context.CreateGroup();
+        var subgroup = context.CreateSubgroup(group.Id);
+        var project = context.CreateProject(subgroup.Id);
+
+        var groups = projectClient.GetGroupsAsync(project.Id, new ProjectGroupsQuery()).ToArray();
+
+        Assert.That(groups.Select(g => g.Id), Contains.Item(group.Id));
+        Assert.That(groups.Select(g => g.Id), Contains.Item(subgroup.Id));
```
louis-z:
```suggestion
        Assert.That(groups.Select(g => g.Id), Is.EquivalentTo(new int[] { group.Id, subgroup.Id }));
```

NGitLab/Impl/ProjectClient.cs
```diff
@@ -173,6 +175,38 @@ public GitLabCollectionResponse<Project> GetForksAsync(string id, ForkedProjectQ
...
+        {
+            url = Utils.AddParameter(url, "shared_min_access_level", (int)query.SharedMinAccessLevel);
+        }
+
+        if (query.SkipGroups is not null)
+        {
+            url = Utils.AddParameter(url, "skip_groups", query.SkipGroups);
+        }
+
+        url = Utils.AddParameter(url, "shared_visible_only", value: query.SharedVisibleOnly);
```
louis-z:
```suggestion
        url = Utils.AddParameter(url, "shared_visible_only", query.SharedVisibleOnly);
```

NGitLab/Models/ProjectGroupsQuery.cs
```diff
@@ -0,0 +1,29 @@
...
+
+    /// <summary>
+    /// Limit to shared groups with at least this access level
+    /// </summary>
+    public AccessLevel? SharedMinAccessLevel { get; set; }
+
+    /// <summary>
+    /// Skip the group IDs passed
+    /// </summary>
+    public int?[] SkipGroups { get; set; }
```
louis-z: Here it's the array that should be optional, not its items, right?

```suggestion
    public int[] SkipGroups { get; set; }
```

louis-z (changes requested):

NGitLab/Models/ProjectGroupsQuery.cs
BallyB: Yes, I couldnt set the entire array nullable so, but sure we can remove it

louis-z (approved):

PR #726, opened 2024-07-22
Add Participant and Unsubscribe functionnalities

tbavard:


NGitLab/IIssueClient.cs
```diff
@@ -180,4 +180,20 @@ public interface IIssueClient
...
     /// <returns></returns>
     Task<Issue> CloneAsync(int projectId, int issueIid, IssueClone issueClone, CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Get Participants
+    /// </summary>
+    /// <param name="projectId">The project id</param>
+    /// <param name="issueIid">The id of the issue in the project's scope.</param>
+    /// <returns>The participants of the issue.</returns>
+    IEnumerable<Participant> GetParticipants(int projectId, int issueIid);
```
Toa741: As they are new methods, you should use `ProjectId` as type instead of regular int.

Then use `projectId.ValueAsUriParameter()` to construct the URLs in the client implementations

NGitLab/Impl/IssueClient.cs
```diff
@@ -192,6 +194,16 @@ public Task<Issue> CloneAsync(int projectId, int issueIid, IssueClone issueClone
         return _api.Post().With(issueClone).ToAsync<Issue>(string.Format(CultureInfo.InvariantCulture, CloneIssueUrl, projectId, issueIid), cancellationToken);
     }
 
+    public IEnumerable<Participant> GetParticipants(int projectId, int issueIid)
```
Toa741: Could you add a test to validate the retrieval of participant and the un-subscription on an issue?

NGitLab/Models/Participant.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System.Text.Json.Serialization;
+
+namespace NGitLab.Models;
+
+public class Participant
+{
+    [JsonPropertyName("id")]
+    public int Id;
```
louis-z: When adding stuff, use properties rather than fields.

louis-z (changes requested):

louis-z (approved):

NGitLab/Impl/IssueClient.cs
tbavard: Lot of stuff is missing to test correctly the subscription. 
I add a test to retrieves participant

PR #736, opened 2024-08-14
Add support for the "Update pipeline metadata" API

ap0llo:
Adds method `UpdateMetadataAsync()` to `IPipelineClient` that calls the ["Update pipeline metadata" API](https://docs.gitlab.com/ee/api/pipelines.html#update-pipeline-metadata).

The REST API currently only supports setting the pipeline name but it looks like it is designed in a way that might allow setting other values in the future, so the metadata to update is passed to `UpdateMetadataAsync()` as an object rather than individual values.
This should allow adding additional data without changing the signature of the method.

### DIFF
--- a/NGitLab.Mock/Clients/PipelineClient.cs
+++ b/NGitLab.Mock/Clients/PipelineClient.cs
@@ -305,4 +305,9 @@ internal sealed class PipelineClient : ClientBase, IPipelineClient
             return Task.FromResult(this[pipelineId]);
         }
     }
+
+    public Task<Models.Pipeline> UpdateMetadataAsync(int pipelineId, PipelineMetadataUpdate update, CancellationToken cancellationToken = default)
+    {
+        throw new NotImplementedException();
+    }
 }
--- a/NGitLab.Mock/Clients/PipelineClient.cs
+++ b/NGitLab.Mock/Clients/PipelineClient.cs
@@ -313,7 +313,7 @@ internal sealed class PipelineClient : ClientBase, IPipelineClient
             var project = GetProject(_projectId, ProjectPermission.Edit);
             var pipeline = project.Pipelines.GetById(pipelineId);
 
-            // Currently, "name" is the only field that can be update dvia the "Update pipeline metadata" API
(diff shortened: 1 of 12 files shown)


NGitLab.Mock/Clients/PipelineClient.cs
```diff
@@ -305,4 +305,28 @@ public GitLabCollectionResponse<PipelineVariable> GetVariablesAsync(int pipeline
...
     }
+
+    public Task<Models.Pipeline> UpdateMetadataAsync(int pipelineId, PipelineMetadataUpdate update, CancellationToken cancellationToken = default)
+    {
+        using (Context.BeginOperationScope())
+        {
+            var project = GetProject(_projectId, ProjectPermission.Edit);
+            var pipeline = project.Pipelines.GetById(pipelineId);
+
+            // Currently, "name" is the only field that can be update dvia the "Update pipeline metadata" API
```
louis-z:
```suggestion
            // Currently, "name" is the only field that can be updated via the "Update pipeline metadata" API
```

NGitLab/Models/PipelineBasic.cs
```diff
@@ -33,4 +33,7 @@ public class PipelineBasic
 
     [JsonPropertyName("web_url")]
     public string WebUrl;
+
+    [JsonPropertyName("name")]
+    public string Name;
```
louis-z: Make this a property instead.

Granted, legacy code uses public fields, but we shouldn't add any more.

NGitLab/Models/Pipeline.cs
```diff
@@ -10,6 +10,9 @@ public class Pipeline
     [JsonPropertyName("id")]
     public int Id;
 
+    [JsonPropertyName("name")]
+    public string Name;
```
louis-z: Make this a property instead.

louis-z (changes requested):
Thanks for your contribution, @ap0llo.

ap0llo:
Thanks for your review!

I have incorporated your feedback

louis-z (approved):

PR #744, opened 2024-08-30
RepositoryClient and CommitCreate updates

jasonwoods-7:
* fixing raw blob api [endpoint](https://docs.gitlab.com/ee/api/repositories.html#raw-blob-content)
* adding support for creating a commit based upon a commit sha
* adding support for setting the executable flag on a file in a new commit

### DIFF
--- a/NGitLab.Tests/CommitsTests.cs
+++ b/NGitLab.Tests/CommitsTests.cs
@@ -117,4 +117,32 @@ public class CommitsTests
         var latestCommit = commitClient.GetCommit(project.DefaultBranch);
         Assert.That(latestCommit.Id, Is.EqualTo(cherryPickedCommit.Id));
     }
+
+    [Test]
+    public async Task Test_commit_can_be_created_from_sha()
+    {
+        using var context = await GitLabTestContext.CreateAsync();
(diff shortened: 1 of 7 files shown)


NGitLab.Tests/CommitsTests.cs
```diff
@@ -117,4 +117,69 @@ public async Task Test_can_cherry_pick_commit()
...
+        var commitClient = context.Client.GetCommits(project.Id);
+
+        var commit = commitClient.GetCommit("main");
+        Assert.That(commit, Is.Not.Null);
+
+        var newCommit = commitClient.Create(new CommitCreate
+        {
+            Branch = "test-start-sha",
+            CommitMessage = "New commit",
+            StartSha = commit.Id.ToString().ToLower(),
```
louis-z:
```suggestion
            StartSha = commit.Id.ToString().ToLowerInvariant(),
```

louis-z (changes requested):

NGitLab.Tests/CommitsTests.cs
jasonwoods-7: updated

louis-z (approved):

PR #745, opened 2024-08-30
Implement asynchronous support for Create, Update, Delete, and FileExists operations

ingshak:
Added CreateAsync, UpdateAsync, DeleteAsync, and FileExistsAsync methods to support asynchronous operations.

### DIFF
--- a/NGitLab/Impl/FilesClient.cs
+++ b/NGitLab/Impl/FilesClient.cs
@@ -70,30 +70,23 @@ public class FilesClient : IFilesClient
         }
     }
 
-    public Task<bool> FileExistsAsync(string filePath, string @ref, CancellationToken cancellationToken = default)
+    public async Task<bool> FileExistsAsync(string filePath, string @ref, CancellationToken cancellationToken = default)
     {
-        return _api.Head().ExecuteAsync(_repoPath + $"/files/{EncodeFilePath(filePath)}?ref={@ref}", cancellationToken)
-            .ContinueWith(t =>
-            {
-                if (t.Status == TaskStatus.RanToCompletion)
-                {
-                    return true;
-                }
-                if (t.Status == TaskStatus.Canceled)
-                {
-                    return false;
-                }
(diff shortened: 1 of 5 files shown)


NGitLab/Impl/FilesClient.cs
```diff
@@ -55,6 +70,32 @@ public bool FileExists(string filePath, string @ref)
         }
     }
 
+    public Task<bool> FileExistsAsync(string filePath, string @ref, CancellationToken cancellationToken = default)
+    {
+        return _api.Head().ExecuteAsync(_repoPath + $"/files/{EncodeFilePath(filePath)}?ref={@ref}", cancellationToken)
+            .ContinueWith(t =>
```
louis-z: Please use `async`/`await`, as it make the code simpler and clearer.

louis-z (changes requested):

NGitLab/Impl/FilesClient.cs
ingshak: Done. Thanks for taking a look!

NGitLab/Impl/FilesClient.cs
```diff
@@ -55,6 +70,25 @@ public bool FileExists(string filePath, string @ref)
...
+        catch (Exception ex) when (ex is OperationCanceledException ||
+                                  (ex is GitLabException gitLabException && gitLabException.StatusCode == HttpStatusCode.NotFound))
+
+        {
+            return false;
+        }
+        catch (Exception)
+        {
+            throw;
+        }
```
louis-z: I think you can do the same as in the synchronous method:

```suggestion
        catch (GitLabException e) when (e.StatusCode == HttpStatusCode.NotFound)
        {
            return false;
        }
```

louis-z (changes requested):

NGitLab/Impl/FilesClient.cs
ingshak: That's fair. Updated! Let me know.

PR #748, opened 2024-09-04
feat: commit, repo, and merge request mock updates

jasonwoods-7:
* use project id when target project id is not specified on a new merge request
* implement get raw blob on repository client
* create action handler should create directory if it does not exist
* throw `GitLabNotFoundException` when directory does not exist to replicate what gitlab does
* add id and mode when retrieving tree items
* use default user when commit author & email not specified


NGitLab.Mock/Repository.cs
```diff
@@ -204,12 +204,12 @@ public Commit Commit(CommitCreate commitCreate)
         var mustCreateBranch = !repo.Commits.Any();
         if (!mustCreateBranch)
         {
-            Commands.Checkout(repo, commitCreate.Branch);
+            Commands.Checkout(repo, commitCreate.StartBranch ?? commitCreate.StartSha);
```
louis-z: There are problems with this change. But to be fair, there were already issues with the existing mock code...

I just added a few more commit creation tests in the main branch, to document GitLab's behavior. We should try to reproduce that behavior in the mocks as much as possible. @jasonwoods-7, could you rebase your branch, look up the aforementioned tests and adapt your PR accordingly? Similar tests should be added for the mocks.

louis-z (changes requested):

NGitLab.Mock/Repository.cs
jasonwoods-7: ok, i think i have all the cases covered now. let me know if i missed any.

NGitLab.Mock/Repository.cs
```diff
@@ -558,14 +584,27 @@ internal string FetchBranchFromFork(Project fork, string branch)
         return $"remotes/{forkRemote.Name}/{branch}";
     }
 
-    private static Tree GetTreeItem(string repositoryPath, string filePath)
+    private Tree GetTreeItem(string repositoryPath, string filePath)
```
jasonwoods-7: I'm working on refactoring this method to more accurately match what GitLab does.

jasonwoods-7: done

NGitLab.Mock.Tests/UsingVerify/RepositoryMockTests.cs
```diff
@@ -0,0 +1,148 @@
+﻿using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using NGitLab.Mock.Clients;
+using NGitLab.Mock.Config;
+using NUnit.Framework;
+using RepositoryGetTreeOptions = NGitLab.Models.RepositoryGetTreeOptions;
+using NThrows = NUnit.Framework.Throws;
```
louis-z: You should be able to use the following alias instead:

```cs
using Throws = NUnit.Framework.Throws;
```

NGitLab.Mock/Repository.cs
```diff
@@ -201,15 +201,35 @@ public Commit Commit(User user, string message, string targetBranch, IEnumerable
...
-        var mustCreateBranch = !repo.Commits.Any();
-        if (!mustCreateBranch)
+        var branchExists = repo.Branches.Any(b => string.Equals(b.FriendlyName, commitCreate.Branch, StringComparison.Ordinal));
+
+        if (branchExists)
+        {
+            throw new GitLabBadRequestException($"A branch called '{commitCreate.Branch}' already exists.");
+        }
+
+        var isBranchEmpty = !repo.Commits.Any();
```
louis-z: It seems to me `isRepoEmpty` would be a more appropriate name here.

louis-z (changes requested):

NGitLab.Mock.Tests/UsingVerify/RepositoryMockTests.cs
jasonwoods-7: updated

NGitLab.Mock/Repository.cs
jasonwoods-7: updated

PR #754, opened 2024-09-09
Add Page property to control pagination. Add asynchronous support for GetCommits, GetCommit, and GetCommitDiff

ingshak:
Added a `Page` property to control pagination, helping to avoid unnecessary data fetches.
Introduced asynchronous support for the following methods: GetCommits, GetCommit, and GetCommitDiff.

### DIFF
--- a/NGitLab.Tests/RepositoryClient/RepositoryClientTests.cs
+++ b/NGitLab.Tests/RepositoryClient/RepositoryClientTests.cs
@@ -1,11 +1,13 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Collections.Specialized;
 using System.Formats.Tar;
 using System.Globalization;
 using System.IO;
 using System.IO.Compression;
 using System.Linq;
 using System.Threading.Tasks;
+using System.Web;
 using NGitLab.Models;
 using NGitLab.Tests.Docker;
 using NUnit.Framework;
@@ -58,6 +60,20 @@ public class RepositoryClientTests
 
     [Test]
     [NGitLabRetry]
+    public async Task GetCommitBySha1Async()
+    {
+        using var context = await RepositoryClientTestsContext.CreateAsync(commitCount: 2);
+
+        var sha1 = context.Commits[0].Id;
+        var commit = await context.RepositoryClient.GetCommitAsync(sha1);
+        Assert.That(commit.Id, Is.EqualTo(sha1));
+        Assert.That(commit.Message, Is.EqualTo(context.Commits[0].Message));
+        Assert.That(commit.WebUrl, Is.EqualTo(context.Commits[0].WebUrl));
+        Assert.That(commit.WebUrl, Is.Not.Null);
+    }
+
+    [Test]
+    [NGitLabRetry]
     public async Task GetCommitBySha1Range()
     {
         using var context = await RepositoryClientTestsContext.CreateAsync(commitCount: 5);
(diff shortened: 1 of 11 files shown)


NGitLab.Tests/RepositoryClient/RepositoryClientTests.cs
```diff
@@ -99,6 +137,35 @@ public async Task GetCommitsSince()
...
 
+    [Test]
+    [NGitLabRetry]
+    public async Task GetCommitsSinceAsync()
+    {
+        // Arrange
+        using var context = await RepositoryClientTestsContext.CreateAsync(commitCount: 5);
+
+        var defaultBranch = context.Project.DefaultBranch;
+        var since = DateTime.UtcNow;
```
louis-z, line 148: Unless I'm missing something, the test here will collect no commit whatsoever. To truly test this, you should create a few (e.g. 5) commits over a small period of time, say one every 200 ms, and then list commits from the creation time of one of the middle commits.

NGitLab.Tests/RepositoryClient/RepositoryClientTests.cs
```diff
@@ -99,6 +137,35 @@ public async Task GetCommitsSince()
...
+        List<Commit> commits = new();
+        await foreach (var commit in context.RepositoryClient.GetCommitsAsync(commitRequest))
+        {
+            commits.Add(commit);
+        }
+
+        // Assert
+        var lastRequestQueryString = context.Context.LastRequest.RequestUri.Query;
+
+        Assert.That(lastRequestQueryString, Does.Contain($"since={expectedSinceValue}"));
```
louis-z, line 166: This assertion has little value. What's more relevant is that we do get only the commits from `since`...

NGitLab.Tests/RepositoryClient/RepositoryClientTests.cs
```diff
@@ -122,6 +189,33 @@ public async Task GetCommitsDoesntIncludeSinceWhenNotSpecified()
...
+        List<Commit> commits = new();
+        await foreach (var commit in context.RepositoryClient.GetCommitsAsync(commitRequest))
+        {
+            commits.Add(commit);
+        }
+
+        // Assert
+        var lastRequestQueryString = context.Context.LastRequest.RequestUri.Query;
+
+        Assert.That(lastRequestQueryString, Does.Not.Contain("since="));
```
louis-z, line 216: Again, I'm not sure this assertion is relevant... I would probably remove that test altogether.

NGitLab.Tests/RepositoryClient/RepositoryClientTests.cs
```diff
@@ -147,6 +241,35 @@ public async Task GetCommitsUntil()
...
     }
 
+    [Test]
+    [NGitLabRetry]
+    public async Task GetCommitsUntilAsync()
+    {
+        // Arrange
+        using var context = await RepositoryClientTestsContext.CreateAsync(commitCount: 5);
+
+        var defaultBranch = context.Project.DefaultBranch;
```
louis-z, line 251: Again, I would stagger the creation commits over a short period of time and validate that specifying `until` works as expected.

NGitLab.Tests/RepositoryClient/RepositoryClientTests.cs
```diff
@@ -170,6 +293,178 @@ public async Task GetCommitsDoesntIncludeUntilWhenNotSpecified()
         Assert.That(lastRequestQueryString, Does.Not.Contain("until="));
     }
 
+    [Test]
+    [NGitLabRetry]
+    public async Task GetCommitsDoesntIncludeUntilWhenNotSpecifiedAsync()
```
louis-z, line 298: Remove test?

NGitLab.Tests/RepositoryClient/RepositoryClientTests.cs
```diff
@@ -170,6 +293,178 @@ public async Task GetCommitsDoesntIncludeUntilWhenNotSpecified()
...
+        };
+
+        // Act
+        var commits = context.RepositoryClient.GetCommits(commitRequest).ToArray();
+
+        // Assert
+        var lastRequestQueryString = context.Context.LastRequest.RequestUri.Query;
+
+        NameValueCollection queryParameters = HttpUtility.ParseQueryString(lastRequestQueryString);
+        Assert.That(queryParameters, Does.Not.Contain("page"));
```
louis-z, line 344: I would prefer that the behavior, from a user's perspective, be tested. With a 5-commit setup and `per_page=2`  you could test what happens when `page` 
- is not set
- is < 3
- is >= 3

NGitLab.Tests/RepositoryClient/RepositoryClientTests.cs
```diff
@@ -170,6 +293,178 @@ public async Task GetCommitsDoesntIncludeUntilWhenNotSpecified()
...
+    }
+
+    [Test]
+    [NGitLabRetry]
+    public async Task GetCommitsDoesntIncludePageWhenNotSpecifiedAsync()
```
louis-z, line 349: Remove test?

louis-z (changes requested):

NGitLab.Tests/RepositoryClient/RepositoryClientTests.cs
ingshak: This function is similar to GetCommitsSince. The common code has been refactored and is now shared between both functions. Should I also update GetCommitsSince to align with these changes? Also, I agree I need to at least add a new test to ensure it is thoroughly tested.

PR #773, opened 2024-10-24
Implement Pipeline Schedule Client

mlemieuxlafontaine-ubi:
Implement the [Pipeline Schedules](https://docs.gitlab.com/ee/api/pipeline_schedules.html) client, to retrieve information about pipeline schedule

### DIFF
--- a/NGitLab.Mock/Clients/PipelineScheduleClient.cs
+++ b/NGitLab.Mock/Clients/PipelineScheduleClient.cs
@@ -44,8 +44,21 @@ internal class PipelineScheduleClient : ClientBase, IPipelineScheduleClient
         }
     }
 
-    public GitLabCollectionResponse<PipelineScheduleBasic> GetAllAsync(CancellationToken cancellationToken = default)
+    public GitLabCollectionResponse<PipelineScheduleBasic> GetAllAsync()
         => GitLabCollectionResponse.Create(All);
+
+    public GitLabCollectionResponse<PipelineBasic> GetAllSchedulePipelines(int id)
+    {
+        using (Context.BeginOperationScope())
+        {
+            var project = GetProject(_projectId, ProjectPermission.View);
+            var pipelines = project.Pipelines
+                .Where(p => string.Equals(p.Source, "schedule", System.StringComparison.Ordinal))
+                .Select(p => p.ToPipelineBasicClient());
+
+            return GitLabCollectionResponse.Create(pipelines);
+        }
+    }
 
     public async Task<Models.PipelineSchedule> GetByIdAsync(int id, CancellationToken cancellationToken = default)
     {
--- a/NGitLab.Mock/Clients/PipelineScheduleClient.cs
+++ b/NGitLab.Mock/Clients/PipelineScheduleClient.cs
@@ -6,6 +6,7 @@ using NGitLab.Mock.Internals;
 using NGitLab.Models;
 
 namespace NGitLab.Mock.Clients;
+
 internal class PipelineScheduleClient : ClientBase, IPipelineScheduleClient
 {
     private readonly ProjectId _projectId;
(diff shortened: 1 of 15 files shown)


NGitLab.Mock/Clients/PipelineScheduleClient.cs
```diff
@@ -0,0 +1,69 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using NGitLab.Mock.Internals;
+using NGitLab.Models;
+
+namespace NGitLab.Mock.Clients;
+internal class PipelineScheduleClient : ClientBase, IPipelineScheduleClient
```
louis-z: Insert an empty line in between.

NGitLab/IPipelineScheduleClient.cs
```diff
@@ -0,0 +1,43 @@
...
+
+public interface IPipelineScheduleClient
+{
+    /// <summary>
+    /// All the scheduled pipelines of the project
+    /// </summary>
+    IEnumerable<PipelineScheduleBasic> All { get; }
+
+    /// <summary>
+    /// Return the details of single schedule
```
louis-z: This is a property, so better not start the description with a verb (mostly reserved for methods).

```suggestion
    /// Details for a single schedule
```

NGitLab/IPipelineScheduleClient.cs
```diff
@@ -0,0 +1,43 @@
...
+
+    /// <summary>
+    /// All the scheduled pipelines of the project
+    /// </summary>
+    /// <param name="cancellationToken"></param>
+    /// <returns></returns>
+    GitLabCollectionResponse<PipelineScheduleBasic> GetAllAsync();
+
+    /// <summary>
+    /// Get all pipelines triggered by a pipeline schedule in a project.
```
louis-z: When documenting a method, use the 3rd person (*descriptive form*, such as `gets`) instead of the 2nd person (*prescriptive form*, such as `get`).

This is a widespread standard, in many languages and in various environments. For example, refer to the following:
- [An example from Microsoft doc](https://docs.microsoft.com/en-us/dotnet/api/system.io.file.getattributes?view=net-6.0#definition)
- [This Google doc](https://developers.google.com/style/reference-verbs)

NGitLab/Impl/PipelineScheduleClient.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    private readonly API _api;
+
+    private readonly string _projectPath;
+    private readonly string _schedulesPath;
+
+    public PipelineScheduleClient(API api, ProjectId projectId)
+    {
+        _api = api;
+        _projectPath = $"{Project.Url}/{projectId.ValueAsUriParameter()}";
```
louis-z: Why have this field? It's only used at construction. You can make it a local variable or inline its value when setting the `_schedulesPath` field.

NGitLab/Models/PipelineSchedule.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text.Json.Serialization;
+
+namespace NGitLab.Models;
+
+public class PipelineSchedule
```
louis-z: Couldn't this class derive from `PipelineScheduleBasic`?

louis-z (changes requested):

NGitLab/Impl/PipelineScheduleClient.cs
mlemieuxlafontaine-ubi: I think I thought I was going to need it at some point in the method, and then forgot to clean that up.

Removed

louis-z (approved):

PR #798, opened 2024-11-19
Be able to search Tags

jsamouh:
Missing search parameter from gitlab API on tag endpoint

https://docs.gitlab.com/ee/api/tags.html

### DIFF
--- a/NGitLab.Tests/TagTests.cs
+++ b/NGitLab.Tests/TagTests.cs
@@ -38,6 +38,11 @@ public class TagTests
     [TestCase("^v0", 2)]
     [TestCase("^v1", 0)]
     [TestCase("v1", 0)]
+    [TestCase("0.5$", 1)]
+    [TestCase("0\\.", 0)]
+    [TestCase(".5$", 1)]
+    [TestCase("\\.5$", 0)]
+    [TestCase(".[0-9]$", 0)]
     public async Task SearchTags(string search, int expectedCount)
     {
         // Arrange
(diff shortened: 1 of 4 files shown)


NGitLab.Tests/TagTests.cs
```diff
@@ -33,6 +33,36 @@ public async Task Test_can_tag_a_project()
         Assert.That(tagsClient.All.FirstOrDefault(x => string.Equals(x.Name, "v0.5", StringComparison.Ordinal)), Is.Null);
     }
 
+    [NGitLabRetry]
+    [TestCase("^v0.5", 1)]
+    [TestCase("^v0", 2)]
+    [TestCase("^v1", 0)]
+    [TestCase("v1", 0)]
```
louis-z, line 40: According to the [GitLab Tags API documentation](https://docs.gitlab.com/ee/api/tags.html#list-project-repository-tags):

| Attribute | Description |
| ---------- | ------------ |
| `search` | Return a list of tags matching the search criteria. You can use `^term` and `term$` to find tags that begin and end with `term`. No other regular expressions are supported. |

Can we validate (and document at the same time) a few more uses and see exactly which ones succeed? For instance, the following ones: `v0`, `0\.`, `.5$`, `\.5$`, `.[0-9]$`

louis-z (changes requested):

NGitLab.Tests/TagTests.cs
jsamouh: added more tests

jsamouh:
thks a lot !

PR #805, opened 2024-12-02
Downgrade "Microsoft.*" and "System.*" packages to v8

louis-z:
- Keep `Microsoft.*` and `System.*` NuGet package versions to 8 for NGitLab.csproj
- Disable their auto-updates


Toa741:
Note: primarily due to the usages on projects that are still on .Net6, some v9 dependencies are not guaranteed to work with it

PR #806, opened 2024-12-03
Add issue references to Issue model

avkranenburg:
This adds the short, relative and full reference for an issue.

### DIFF
--- a/NGitLab/Models/References.cs
+++ b/NGitLab/Models/References.cs
@@ -5,11 +5,11 @@ namespace NGitLab.Models;
 public class References
 {
     [JsonPropertyName("short")]
-    public string Short;
+    public string Short { get; set; }
 
     [JsonPropertyName("relative")]
-    public string Relative;
+    public string Relative { get; set; }
 
     [JsonPropertyName("full")]
-    public string Full;
+    public string Full { get; set; }
 }
--- a/NGitLab/Models/References.cs
+++ b/NGitLab/Models/References.cs
@@ -1,0 +1,15 @@
+﻿using System.Text.Json.Serialization;
+
+namespace NGitLab.Models;
+
+public class References
+{
+    [JsonPropertyName("short")]
+    public string Short;
+
+    [JsonPropertyName("relative")]
+    public string Relative;
+
+    [JsonPropertyName("full")]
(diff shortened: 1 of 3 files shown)


NGitLab/Models/References.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System.Text.Json.Serialization;
+
+namespace NGitLab.Models;
+
+public class References
+{
+    [JsonPropertyName("short")]
+    public string Short;
```
louis-z: Please define these public fields as properties instead.

louis-z (changes requested):

NGitLab/Models/References.cs
avkranenburg: Done, but is there a reason why other classes like User, Issue, Author, Assignee and Milestone are not using properties? (I used fields to be consistent)

louis-z: Older types use fields but when introducing new types we try to enforce use of properties.

NGitLab/PublicAPI.Unshipped.txt
```diff
@@ -2117,6 +2117,7 @@ NGitLab.Models.Issue.MergeRequestsCount.get -> int
 NGitLab.Models.Issue.MergeRequestsCount.set -> void
 NGitLab.Models.Issue.Milestone -> NGitLab.Models.Milestone
 NGitLab.Models.Issue.MovedToId -> int?
+NGitLab.Models.Issue.References -> NGitLab.Models.References
```
louis-z: @avkranenburg, you're still missing info in this file. Also, this line is not at the correct location; it should be right after the next one, to remain sorted.

If you're using Visual Studio, I would recommend that you
- Undo all changes to the file (i.e. bring it back to the state it's in in the `main` branch)
- Use the IDE's embedded refactoring and select "Fix all occurrences in: **Solution**"

image

If you're using VS Code, Rider or something else, I'm not sure how to achieve the same effect.

avkranenburg: I don't use it often but do have it installed, will use that!

louis-z (approved):

PR #810, opened 2024-12-13
Event's Id sould be long type

stevenxi:
We got json deserialize exception when loading events for an user.

As the 'Id' field overflows integer. So it has to be long.


louis-z:
Hi @stevenxi,

Thanks for your contribution, but your changes entail a breaking API change so we can't currently accept this PR.

Note however that we're currently planning on converting most (if not all) `int` IDs to `long` and subsequently bumping the major version. Refer to #491.

PR #825, opened 2025-01-07
Make sure GetTree() in mock reproduces behavior on GitLab

DomLatr:


NGitLab.Mock.Tests/UsingVerify/RepositoryMockTests.cs
```diff
@@ -26,10 +24,10 @@ public void Test_get_tree_when_subdirectory_does_not_exist_throws_not_found()
         var repository = client.GetRepository(1);
 
         // Act
-        var handler = () => repository.GetTreeAsync(new RepositoryGetTreeOptions { Path = "non-existing-directory" });
+        var tree = repository.GetTreeAsync(new RepositoryGetTreeOptions { Path = "non-existing-directory" });
```
louis-z: The `GetTreeAsync` method is not async, but returns a `GitLabCollectionResponse<Tree>`... which can then be enumerated synchronously or asynchronously.

(The method naming is unfortunate as it suggests the call can be awaited. We should probably revisit that name in an ulterior major version, but that's beside the point here. 😁)

Perhaps renaming `tree` to `treeCollection` might be a good idea. What do you think?

NGitLab.Mock/Repository.cs
```diff
@@ -552,6 +552,11 @@ internal IEnumerable<Tree> GetTree(RepositoryGetTreeOptions repositoryGetTreeOpt
 
         if (tree is null)
         {
+            // Based on test GetAllTreeObjectsAtInvalidPath(), default behavior for
+            // GetTree() when Path doesn't exists should be to return empty
```
louis-z:
```suggestion
            // As confirmed by the GetAllTreeObjectsAtInvalidPath() test, GitLab's behavior for
            // GetTree() when Path doesn't exist is to return an empty collection.
```

louis-z (changes requested):

louis-z (approved):

PR #828, opened 2025-01-13
Add AutoStopAt to EnvironmentInfo

alex-hempel:
This PR adds AutoStopAt to EnvironmentInfo. 

GitLab environments have a property `auto_stop_at` which is not yet part of EnvironmentInfo in this client.

The file NGitLab/PublicAPI.Unshipped.txt has been updated as well to allow for the new property.

### DIFF
--- a/NGitLab/Models/EnvironmentInfo.cs
+++ b/NGitLab/Models/EnvironmentInfo.cs
@@ -1,4 +1,5 @@
-﻿using System.Text.Json.Serialization;
+﻿using System;
+using System.Text.Json.Serialization;
 
 namespace NGitLab.Models;
 
@@ -21,4 +22,7 @@ public class EnvironmentInfo
 
     [JsonPropertyName("last_deployment")]
     public EnvironmentLastDeployment LastDeployment { get; set; }
+
+    [JsonPropertyName("auto_stop_at")]
+    public DateTime AutoStopAt { get; set;  }
 }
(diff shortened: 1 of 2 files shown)


NGitLab/Models/EnvironmentInfo.cs
```diff
@@ -21,4 +22,7 @@ public class EnvironmentInfo
 
     [JsonPropertyName("last_deployment")]
     public EnvironmentLastDeployment LastDeployment { get; set; }
+
+    [JsonPropertyName("auto_stop_at")]
+    public DateTime AutoStopAt { get; set;  }
```
Toa741: Not very clear from the doc or code from gitlab, but I suppose it could be null? If so, put it as nullable

alex-hempel: Yes, it could be null. I tested with a local build and it works either way, but if this is the preference, I will set it as nullable.

Toa741: I think so, although there are probably a lot of places datetimes are nullable, but not noted as such in NGitLab, I would say that for new additions, it would be preferable using `DateTime?`.

Ok for you @louis-z ?

Toa741:
```suggestion
    public DateTime? AutoStopAt { get; set;  }
```

PR #831, opened 2025-01-15
Add route to get project templates and merge request templates

Codex04:


NGitLab/Impl/ProjectTemplateType.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System.Runtime.Serialization;
+
+namespace NGitLab.Impl;
+
+public enum ProjectTemplateType
+{
+    [EnumMember(Value = "dockerfiles")]
+    Dockerfiles,
```
ivan-pavlik:
```suggestion
    DockerFiles,
```

NGitLab/Impl/ProjectTemplateType.cs
```diff
@@ -0,0 +1,24 @@
...
+
+    [EnumMember(Value = "gitignores")]
+    Gitignores,
```
ivan-pavlik:
```suggestion
    GitIgnores,
```

NGitLab/Impl/ProjectTemplateType.cs
ivan-pavlik, line 1: Shouldn't this be put into 'Models' subfolder?

NGitLab/Impl/ProjectTemplateType.cs
Codex04: Dockerfile is written in one word. It doesn't make sense to put a capital F.

NGitLab/Impl/ProjectTemplateType.cs
Codex04: Same for Gitignore that is written .gitignore

NGitLab/Impl/ProjectTemplateType.cs
Codex04: I'll do the move

PR #832, opened 2025-01-16
Add endpoint to subscribe to the issue

ivan-pavlik:
Adding this endpoint
https://docs.gitlab.com/ee/api/issues.html#subscribe-to-an-issue

Similar to unsubscribe from the issue endpoint.

### DIFF
--- a/NGitLab/IIssueClient.cs
+++ b/NGitLab/IIssueClient.cs
@@ -190,7 +190,7 @@ public interface IIssueClient
     IEnumerable<Participant> GetParticipants(ProjectId projectId, long issueIid);
 
     /// <summary>
-    /// Subscribes to the issue.
+    /// Subscribes the authenticated user to an issue to receive notifications.
     /// </summary>
     /// <param name="projectId">The project id</param>
     /// <param name="issueIid">The id of the issue in the project's scope.</param>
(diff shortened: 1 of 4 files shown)


NGitLab/IIssueClient.cs
```diff
@@ -190,7 +190,15 @@ public interface IIssueClient
     IEnumerable<Participant> GetParticipants(ProjectId projectId, long issueIid);
 
     /// <summary>
-    /// Unsubscribe to the issue
+    /// Subscribe to the issue.
```
louis-z: When documenting a method, use the 3rd person (*descriptive form*, such as `gets`) instead of the 2nd person (*prescriptive form*, such as `get`).

This is a widespread standard, in many languages and in various environments. For example, refer to the following:
- [An example from Microsoft doc](https://learn.microsoft.com/en-us/dotnet/api/system.io.file.getattributes?view=net-9.0#system-io-file-getattributes(microsoft-win32-safehandles-safefilehandle))
- [Google developer documentation style guide](https://developers.google.com/style/reference-verbs)

```suggestion
    /// Subscribes the authenticated user to an issue to receive notifications.
```

louis-z (changes requested):

NGitLab/IIssueClient.cs
ivan-pavlik: Applied.

PR #837, opened 2025-01-20
Add SHA256 to FileData

qiao9580:


NGitLab/Models/FileData.cs
```diff
@@ -21,6 +21,9 @@ public class FileData
     [JsonPropertyName("content")]
     public string Content { get; set; }
 
+    [JsonPropertyName("content_sha256")]
+    public string ContentSHA256 { get; set; }
```
louis-z: We usually name our "SHA" properties, fields, variables, etc. with a capital 'S' followed by lowercase "ha":

```suggestion
    public string ContentSha256 { get; set; }
```

louis-z (changes requested):

louis-z (approved):

Issue #841, opened 2025-02-03
NGitLab.Mocks should throw same exceptions as NGitLab


louis-z:
The mocks are meant to fake NGitLab as faithfully as possible.  As such, they should throw the same exceptions as their true NGitLab counterparts.

PR #842, opened 2025-02-03
Adapt mocks a bit so that they behave more like NGitLab

louis-z:
- Get rid of Mocks-specific `GitLabBadRequestException`, `GitLabForbiddenException` and `GitLabNotFoundException`
- Use NGitLab's `GitLabException` instead
- Add internal static methods to `GitLabException`, to ease the creation of specific exceptions

Closes to #841


Toa741:
- NotFound exception is only in mock
- Can be misleading (catching this type only instead of the base exception)
- Remove `GitLabNotFoundException`, use ctors or methods to create GitLabException with correct status code, so the mock & client always return the same type of exception

NGitLab.Mock/Clients/ClientBase.cs
```diff
@@ -85,8 +85,10 @@ protected Project GetProject(object id, ProjectPermission permissions)
             _ => throw new ArgumentException($"Id of type '{id.GetType()}' is not supported"),
         };
 
-        if (project == null || !project.CanUserViewProject(Context.User))
-            throw new GitLabNotFoundException("Project does not exist or User doesn't have permission to view it");
+        if (project is null || !project.CanUserViewProject(Context.User))
+        {
+            throw GitLabException.NotFound();
```
Toa741: Intended to remove the message?

louis-z: Oops, my bad. Will put it back. Good catch!

Issue #840, opened 2025-02-03
Why is the token mandatory?


MrArturGar:
I recently wanted to add `NGitLab `to my project but ran into an issue where the `GitLabCredentials` constructor enforces a mandatory API token:

```
public GitLabCredentials(string hostUrl, string apiToken)
{
    if (string.IsNullOrEmpty(hostUrl))
        throw new ArgumentException("HostUrl is mandatory", nameof(hostUrl));
    if (string.IsNullOrEmpty(apiToken))
        throw new ArgumentException("Token is mandatory", nameof(apiToken));

    ValidateHostUrl(hostUrl);

    HostUrl = GetApiUrl(hostUrl);
    _apiToken = apiToken;
}
```
However, looking at this condition in `HttpRequestor.GitLabRequest.cs`, it seems that requests without a token are actually supported:
[GitLabRequest.cs#L51](https://github.com/ubisoft/NGitLab/blob/main/NGitLab/Impl/HttpRequestor.GitLabRequest.cs#L51)

Is this an oversight, or is the token intentionally required in `GitLabCredentials`? If requests can be made without a token, would it make sense to allow `GitLabCredentials` to accept a null or empty token?

Toa741:
Hello,
It's probably some code put in place a long time ago. As most API request requires to be authenticated, maybe it has just been made to correlate that.

Though, it's technically valid that you would want to do some requests without authentication and the option should be available.

If you want, you can contribute to the project and make a pull request to add it. Otherwise, we will try to make the change as soon as we can.

MrArturGar:
I apologize, just now saw the change on my ticket. Thank you so much!

PR #845, opened 2025-02-06
Expose GitLabClient without authentication

Toa741:
- References https://github.com/ubisoft/NGitLab/issues/840
- Add ctors to be able to get a GitLabClient without any authentication
- Didn't touch the actuall logic that, with username/password, set the ApiToken after getting a session

### DIFF
--- a/NGitLab/GitLabClient.cs
+++ b/NGitLab/GitLabClient.cs
@@ -51,7 +51,7 @@ public class GitLabClient : IGitLabClient
     }
 
     /// <summary>
-    /// Initialize a GitLab client without any authentication and default options
+    /// Initialize a GitLab client without any authentication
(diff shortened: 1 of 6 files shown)


NGitLab/GitLabClient.cs
```diff
@@ -50,21 +50,67 @@ public RequestOptions Options
         set => _api.RequestOptions = value;
     }
 
+    /// <summary>
+    /// Initialize a GitLab client without any authentication and default options
```
louis-z:
```suggestion
    /// Initialize a GitLab client without any authentication
```

louis-z (changes requested):

louis-z (approved):

PR #851, opened 2025-02-13
Support query parameters in members and labels searching endpoints

pvgritsenko-ansible:
Support query parameters in GET projects/:id/labels, GET groups/:id/labels, GET projects/:id/members and GET groups/:id/members endpoints.
Also present PR updates the GET projects/:id/users

### DIFF
--- a/NGitLab.Mock/Clients/MembersClient.cs
+++ b/NGitLab.Mock/Clients/MembersClient.cs
@@ -19,10 +19,15 @@ internal sealed class MembersClient : ClientBase, IMembersClient
 
     public IEnumerable<Membership> OfProject(string projectId)
     {
-        return OfProject(projectId, includeInheritedMembers: false);
+        return OfProject(projectId, includeInheritedMembers: false, query: null);
     }
 
-    public IEnumerable<Membership> OfProject(string projectId, bool includeInheritedMembers, MemberQuery query = null)
+    public IEnumerable<Membership> OfProject(string projectId, bool includeInheritedMembers)
+    {
+        return OfProject(projectId, includeInheritedMembers, query: null);
+    }
+
+    public IEnumerable<Membership> OfProject(string projectId, bool includeInheritedMembers, MemberQuery query)
     {
         using (Context.BeginOperationScope())
         {
@@ -30,6 +35,11 @@ internal sealed class MembersClient : ClientBase, IMembersClient
             var members = project.GetEffectivePermissions(includeInheritedMembers).Permissions;
             return members.Select(member => member.ToMembershipClient());
         }
+    }
+
+    public GitLabCollectionResponse<Membership> OfProjectAsync(ProjectId projectId, bool includeInheritedMembers)
+    {
+        return OfProjectAsync(projectId, includeInheritedMembers, query: null);
     }
 
     public GitLabCollectionResponse<Membership> OfProjectAsync(ProjectId projectId, bool includeInheritedMembers = false, MemberQuery query = null)
@@ -133,10 +143,15 @@ internal sealed class MembersClient : ClientBase, IMembersClient
 
     public IEnumerable<Membership> OfGroup(string groupId)
     {
-        return OfGroup(groupId, includeInheritedMembers: false);
+        return OfGroup(groupId, includeInheritedMembers: false, null);
     }
 
-    public IEnumerable<Membership> OfGroup(string groupId, bool includeInheritedMembers, MemberQuery query = null)
+    public IEnumerable<Membership> OfGroup(string groupId, bool includeInheritedMembers)
+    {
+        return OfGroup(groupId, includeInheritedMembers, query: null);
+    }
+
+    public IEnumerable<Membership> OfGroup(string groupId, bool includeInheritedMembers, MemberQuery query)
     {
         using (Context.BeginOperationScope())
         {
@@ -144,6 +159,11 @@ internal sealed class MembersClient : ClientBase, IMembersClient
             var members = group.GetEffectivePermissions(includeInheritedMembers).Permissions;
             return members.Select(member => member.ToMembershipClient());
         }
+    }
+
+    public GitLabCollectionResponse<Membership> OfGroupAsync(GroupId groupId, bool includeInheritedMembers)
+    {
+        return OfGroupAsync(groupId, includeInheritedMembers, query: null);
     }
 
     public GitLabCollectionResponse<Membership> OfGroupAsync(GroupId groupId, bool includeInheritedMembers = false, MemberQuery query = null)
--- a/NGitLab.Mock/Clients/MembersClient.cs
+++ b/NGitLab.Mock/Clients/MembersClient.cs
@@ -22,7 +22,7 @@ internal sealed class MembersClient : ClientBase, IMembersClient
         return OfProject(projectId, includeInheritedMembers: false);
     }
 
-    public IEnumerable<Membership> OfProject(string projectId, bool includeInheritedMembers)
+    public IEnumerable<Membership> OfProject(string projectId, bool includeInheritedMembers, MemberQuery query = null)
     {
         using (Context.BeginOperationScope())
         {
@@ -32,7 +32,7 @@ internal sealed class MembersClient : ClientBase, IMembersClient
         }
     }
 
-    public GitLabCollectionResponse<Membership> OfProjectAsync(ProjectId projectId, bool includeInheritedMembers = false)
+    public GitLabCollectionResponse<Membership> OfProjectAsync(ProjectId projectId, bool includeInheritedMembers = false, MemberQuery query = null)
     {
         return GitLabCollectionResponse.Create(OfProject(projectId.ValueAsString(), includeInheritedMembers));
     }
@@ -136,7 +136,7 @@ internal sealed class MembersClient : ClientBase, IMembersClient
         return OfGroup(groupId, includeInheritedMembers: false);
     }
 
-    public IEnumerable<Membership> OfGroup(string groupId, bool includeInheritedMembers)
+    public IEnumerable<Membership> OfGroup(string groupId, bool includeInheritedMembers, MemberQuery query = null)
     {
         using (Context.BeginOperationScope())
         {
@@ -146,7 +146,7 @@ internal sealed class MembersClient : ClientBase, IMembersClient
         }
     }
 
-    public GitLabCollectionResponse<Membership> OfGroupAsync(GroupId groupId, bool includeInheritedMembers = false)
+    public GitLabCollectionResponse<Membership> OfGroupAsync(GroupId groupId, bool includeInheritedMembers = false, MemberQuery query = null)
     {
         return GitLabCollectionResponse.Create(OfGroup(groupId.ValueAsString(), includeInheritedMembers));
     }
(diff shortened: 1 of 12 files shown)


louis-z:
Thanks for your contribution, @pvgritsenko-ansible.

Adding support for query parameters to improve search capabilities is a good idea. However, by introducing additional method parameters (even if optional) you're breaking the binary compatibility. Consider creating new methods instead of modifying existing ones.

pvgritsenko-ansible:
Thanks for your comments, @louis-z 
Oh, ok. I'm a bit unfamiliar with this ways to resolve the compability issues. I've made additional methods to follow the rules mentioned [here](https://github.com/dotnet/roslyn/blob/main/docs/Adding%20Optional%20Parameters%20in%20Public%20API.md).
May I missed something else?

NGitLab/Models/MemberQuery.cs
```diff
@@ -0,0 +1,36 @@
...
+
+    /// <summary>
+    /// Filter the results on the given user IDs.
+    /// </summary>
+    public List<int> UserIds { get; set; }
+
+    /// <summary>
+    /// Filter skipped users out of the results.
+    /// </summary>
+    public List<int> SkipUsers { get; set; }
```
Toa741: Forgot to add it with AddParameter when constructing the url?

NGitLab.Mock/Clients/MembersClient.cs
```diff
@@ -32,7 +37,12 @@ public IEnumerable<Membership> OfProject(string projectId, bool includeInherited
         }
     }
 
-    public GitLabCollectionResponse<Membership> OfProjectAsync(ProjectId projectId, bool includeInheritedMembers = false)
+    public GitLabCollectionResponse<Membership> OfProjectAsync(ProjectId projectId, bool includeInheritedMembers)
```
Toa741, line 40: I think this one (there are others) would still represent a breaking change.

What I would suggest is:

- don't touch the existing methods' signature. 
- create new ones (the ones you need) without any optional paremeters (ex on the one just below: `public GitLabCollectionResponse<Membership> OfProjectAsync(ProjectId projectId, bool includeInheritedMembers, MemberQuery query)`)

NGitLab.Mock/Clients/MembersClient.cs
```diff
@@ -146,7 +161,12 @@ public IEnumerable<Membership> OfGroup(string groupId, bool includeInheritedMemb
         }
     }
 
-    public GitLabCollectionResponse<Membership> OfGroupAsync(GroupId groupId, bool includeInheritedMembers = false)
+    public GitLabCollectionResponse<Membership> OfGroupAsync(GroupId groupId, bool includeInheritedMembers)
```
Toa741, line 164: As previous comment, don't touch existing signature

NGitLab/IMembersClient.cs
```diff
@@ -15,7 +15,11 @@ public interface IMembersClient
 
     IEnumerable<Membership> OfProject(string projectId, bool includeInheritedMembers);
 
-    GitLabCollectionResponse<Membership> OfProjectAsync(ProjectId projectId, bool includeInheritedMembers = false);
+    IEnumerable<Membership> OfProject(string projectId, bool includeInheritedMembers, MemberQuery query);
```
Toa741, line 18: As previous comment, don't touch existing signature (OfProjectAsync & OfGroupAsync)

NGitLab/Impl/LabelClient.cs
```diff
@@ -21,17 +21,47 @@ public LabelClient(API api)
...
     {
-        return _api.Get().GetAll<Label>(string.Format(CultureInfo.InvariantCulture, ProjectLabelUrl, projectId));
+        return ForProject(projectId, query: null);
+    }
+
+    public IEnumerable<Label> ForProject(long projectId, LabelQuery query)
+    {
+        string url = string.Format(CultureInfo.InvariantCulture, ProjectLabelUrl, projectId);
+
+        if (query != null)
```
Toa741: Can it be added in a method to avoid duplication? Something like

```csharp
url = AddLabelParameterQuery(url, query);

private static string AddLabelParameterQuery(string url, LabelQuery query)
{
    if (query == null)
        return url;
    
    url = Utils.AddParameter(url, "with_counts", query.WithCounts);
    url = Utils.AddParameter(url, "per_page", query.PerPage);
    url = Utils.AddParameter(url, "search", query.Search);
    url = Utils.AddParameter(url, "include_ancestor_groups ", query.IncludeAncestorGroups);

    return url;
}
```

Same thing for `MemberQuery`

NGitLab/Impl/LabelClient.cs
```diff
@@ -21,17 +21,47 @@ public LabelClient(API api)
...
+            url = Utils.AddParameter(url, "include_ancestor_groups ", query.IncludeAncestorGroups);
+        }
+
+        return _api.Get().GetAll<Label>(url);
     }
 
     public Label GetProjectLabel(long projectId, string name)
     {
-        return ForProject(projectId).FirstOrDefault(x => string.Equals(x.Name, name, StringComparison.Ordinal));
+        return ForProject(projectId, new LabelQuery() { Search = name }).FirstOrDefault(x => string.Equals(x.Name, name, StringComparison.Ordinal));
```
Toa741, line 48: It seems there is no existing tests for labels.

- Would you mind creating one for this method? 
- Ideally, also one for the new label query object (no need to test every parameters, just one to make sure that it works properly)
- Also, one test for the new `MemberQuery` query objects with one method it is used in?

If you need assistance for testing in the NGitLab.Tests project, let us know :)

NGitLab/Impl/MembersClient.cs
```diff
@@ -17,41 +17,70 @@ public MembersClient(API api)
...
     }
 
-    private IEnumerable<Membership> GetAll(string url, bool includeInheritedMembers)
+    private IEnumerable<Membership> GetAll(string url, bool includeInheritedMembers, MemberQuery query)
     {
         url += "/members";
         if (includeInheritedMembers)
         {
             url += "/all";
+
```
Toa741:
```suggestion
```

NGitLab.Mock/Clients/MembersClient.cs
pvgritsenko-ansible: Actually I didn't want to modify existing method signature. But if I leave existing method: `GitLabCollectionResponse<Membership> OfProjectAsync(ProjectId projectId, bool includeInheritedMembers = false)`  
as it is and add new signature:
 `GitLabCollectionResponse<Membership> OfProjectAsync(ProjectId projectId, bool includeInheritedMembers, MemberQuery query)` , that you suggested then following warning will apear: 
"'OfProjectAsync' violates the backcompat requirement: 'API with optional parameter(s) should have the most parameters amongst its public overloads.'"

So that's why I've desided to add method with all optional parameters and modify the existing one to have only mandatory parameters.

Toa741: Indeed, after some check, that make sense, will resolve the associated comments

NGitLab/Models/MemberQuery.cs
pvgritsenko-ansible: Oh, yes. Thanks! Fixed.

NGitLab.Mock/Clients/MembersClient.cs
pvgritsenko-ansible: Good news for me=) I was a bit confused at this point yesterday.

NGitLab/Impl/LabelClient.cs
pvgritsenko-ansible: Great idea! Thanks, applied.

NGitLab/Impl/LabelClient.cs
pvgritsenko-ansible: Sorry I didn't plan to add tests. And unfortunally I didn't have much time to do that. We use your library in our enterprise project and we are waiting for these changes. 

If it is necessary I guess I could add tests, but in this case I need your assistance as you said.

NGitLab/Impl/MembersClient.cs
pvgritsenko-ansible: Applied

NGitLab/Impl/LabelClient.cs
Toa741: I will validate the changes, we can bypass the tests

NGitLab/Models/MemberQuery.cs
```diff
@@ -0,0 +1,36 @@
...
+
+    /// <summary>
+    /// Filters results based on a given name, email, or username. Use partial values to widen the scope of the query.
+    /// </summary>
+    public string Query { get; set; }
+
+    /// <summary>
+    /// Filter the results on the given user IDs.
+    /// </summary>
+    public List<int> UserIds { get; set; }
```
Toa741: `UserIds` and `SkipUsers` should be `long[]`.

Will also fix an issue when converting to url parameters (there is a dedicated overload for long[])

NGitLab/Models/LabelQuery.cs
```diff
@@ -0,0 +1,24 @@
...
+
+    /// <summary>
+    /// Include ancestor groups. Defaults to true.
+    /// </summary>
+    public bool? IncludeAncestorGroups { get; set; }
+
+    /// <summary>
+    /// Whether or not to include issue and merge request counts. Defaults to false.
+    /// </summary>
+    public bool? WithCounts { get; set; }
```
Toa741, line 23: Was it intended for you to use?

At the moment, the `Label` class doesn't have the corresponding properties populated when WithCounts is true (open_issues_count, closed_issues_count, open_merge_requests_count)

Either remove the filter, or add the properties in the Label class

NGitLab/Models/MemberQuery.cs
pvgritsenko-ansible: I see, thanks! Fixed

NGitLab/Models/LabelQuery.cs
pvgritsenko-ansible: I wasn't going to use this option and just wanted the model to match the parameters in the documentation. I have expanded the Label class

NGitLab/Models/Label.cs
```diff
@@ -12,4 +12,13 @@ public class Label
...
     public string Description { get; set; }
+
+    [JsonPropertyName("open_issues_count")]
+    public string OpenIssuesCount { get; set; }
+
+    [JsonPropertyName("closed_issues_count")]
+    public string ClosedIssuesCount { get; set; }
+
+    [JsonPropertyName("open_merge_requests_count")]
+    public string OpenMergeRequestsCount { get; set; }
```
Toa741:
```suggestion
    public int OpenIssuesCount { get; set; }

    [JsonPropertyName("closed_issues_count")]
    public int ClosedIssuesCount { get; set; }

    [JsonPropertyName("open_merge_requests_count")]
    public int OpenMergeRequestsCount { get; set; }
```

pvgritsenko-ansible: Sorry for that. Fixed

Toa741:
@pvgritsenko-ansible , can you check the PublicAPI.Unshipped.txt, you probably have some warnings saying it's not up to date (UserIds/SkipUsers)

PR #857, opened 2025-02-18
Add GetMetaData API

qiao9580:
This interface is used to obtain the header of a file without obtaining the body of the file, in order to improve speed


qiao9580:
How can I return content similar to the FileData type in the GetMetaData interface

qiao9580:
The merge has been abandoned and other methods should be used to compare the files

PR #860, opened 2025-02-19
Add ci lint job

PaysPlat:
Hi,
I added support of jobs details in LintCi response when setting the switch IncludeJobs in request


louis-z:
Thanks for your contribution, @PaysPlat.

You'll need to modify `NGitLab\PublicAPI.Unshipped.txt`, to document changes made to the public API. I recommend that you do so using VS2022's "Quick Actions and Refactorings..." rather than modifying the file by hand.

image
 
By selecting "Fix all occurrences in: Solution", it will make all the necessary changes.

PaysPlat:
Thank you for your quick feedback.
Sorry I did not know this way of declaring API, I have fixed it following your indications.

PR #861, opened 2025-02-19
Address code analyzer-reported issues

louis-z:
Address some of the analyzer rule violations
- that could automatically be fixed via the VS2022 IDE, and
- whose fix makes the code simpler, clearer or more performant

### DIFF
--- a/NGitLab.Tests/JobTests.cs
+++ b/NGitLab.Tests/JobTests.cs
@@ -13,29 +13,29 @@ public class JobTests
 {
     internal static void AddGitLabCiFile(IGitLabClient client, Project project, int jobCount = 1, bool manualAction = false, string branch = null, bool pipelineSucceeds = true)
     {
-        var content = @"
-variables:
-  CI_DEBUG_TRACE: ""true""
-";
+        var content = """
+            variables:
+              CI_DEBUG_TRACE: "true"
+            """;
 
         for (var i = 0; i < jobCount; i++)
         {
-            content += $@"
-build{i.ToString(CultureInfo.InvariantCulture)}:
-  script:
-    - echo test
-    - echo test > file{i.ToString(CultureInfo.InvariantCulture)}.txt
-    - exit {(pipelineSucceeds ? "0" : "1")}
-  artifacts:
-    paths:
-      - '*.txt'
-";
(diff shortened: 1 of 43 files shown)


NGitLab.Tests/JobTests.cs
```diff
@@ -13,29 +13,32 @@ public class JobTests
...
     internal static void AddGitLabCiFile(IGitLabClient client, Project project, int jobCount = 1, bool manualAction = false, string branch = null, bool pipelineSucceeds = true)
     {
-        var content = @"
-variables:
-  CI_DEBUG_TRACE: ""true""
-";
+        var content = """
+            variables:
+              CI_DEBUG_TRACE: "true"
+
```
Toa741, line 19:
```suggestion
```

NGitLab.Tests/JobTests.cs
```diff
@@ -13,29 +13,32 @@ public class JobTests
...
+            content += $"""
+                build{i.ToString(CultureInfo.InvariantCulture)}:
+                  script:
+                    - echo test
+                    - echo test > file{i.ToString(CultureInfo.InvariantCulture)}.txt
+                    - exit {(pipelineSucceeds ? "0" : "1")}
+                  artifacts:
+                    paths:
+                      - '*.txt'
+
```
Toa741, line 33:
```suggestion
```

NGitLab.Tests/JobTests.cs
```diff
@@ -13,29 +13,32 @@ public class JobTests
...
+                """;
 
             if (manualAction)
             {
-                content += @"
-  when: manual
-";
+                content += """
+                      when: manual
+
```
Toa741, line 40:
```suggestion
```

NGitLab.Tests/JobTests.cs
louis-z: We need that "empty" line. Without it, subsequent string concatenations will result in invalid YAML.

Refer to [Microsoft's documentation about raw strings](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/raw-string):

`The newline before the closing quotes isn't included in the literal string.`

PR #870, opened 2025-03-10
Add patch and protected branch update

JesseKlaasse:
In my own personal projects, I need the update functionality for protected branches.

First, I tried to implement it in my own projects using extension methods and some reflection, but I ran into some issues, so I decided to create a PR instead. I realize that there are no tests yet, but I thought it would be wise to check if there is broader interest in this first.

### DIFF
--- a/NGitLab.Mock/Clients/ProtectedBranchClient.cs
+++ b/NGitLab.Mock/Clients/ProtectedBranchClient.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using NGitLab.Models;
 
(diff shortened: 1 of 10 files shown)


NGitLab.Mock/Clients/ProtectedBranchClient.cs
```diff
@@ -67,4 +67,108 @@ public void UnprotectBranch(string branchName)
...
+            }
+
+            if (protectedBranchUpdate.AllowForcePush is not null)
+            {
+                protectedBranch.AllowForcePush = protectedBranchUpdate.AllowForcePush.Value;
+            }
+
+            if (protectedBranchUpdate.AllowedToMerge is not null)
+            {
+                foreach (var allowedToMerge in protectedBranchUpdate.AllowedToMerge)
```
Toa741: The logic for AllowedToMerge & AllowedToPush seems quite similar, could it be shared with a method?

NGitLab/PublicAPI.Unshipped.txt
```diff
@@ -1080,6 +1080,7 @@ NGitLab.IProtectedBranchClient.GetProtectedBranch(string branchName) -> NGitLab.
 NGitLab.IProtectedBranchClient.GetProtectedBranches(string search = null) -> NGitLab.Models.ProtectedBranch[]
 NGitLab.IProtectedBranchClient.ProtectBranch(NGitLab.Models.BranchProtect branchProtect) -> NGitLab.Models.ProtectedBranch
 NGitLab.IProtectedBranchClient.UnprotectBranch(string branchName) -> void
+NGitLab.IProtectedBranchClient.UpdateProtectedBranch(string branchName, NGitLab.Models.ProtectedBranchUpdate protectedBranchUpdate) -> NGitLab.Models.ProtectedBranch
```
Toa741, line 1084: You probably miss quite a lot of entries in this file. If you use Visual Studio (not sure for vscode), you could find a new property and in the suggestions, apply on the entire project so it add all the new elements at once

Toa741:
About the tests: there is an existing test on [ProtectedBranches](https://github.com/ubisoft/NGitLab/blob/main/NGitLab.Tests/ProtectedBranchTests.cs).

You could likely make a little one to check the updates. If you don't have docker locally, we can still validate the test on the CI

NGitLab.Mock/Clients/ProtectedBranchClient.cs
JesseKlaasse: @Toa741 You're right. I have solved this in a new commit.

NGitLab/PublicAPI.Unshipped.txt
JesseKlaasse: @Toa741 I have fixed this now. By hand, using Rider ;)

JesseKlaasse:
> About the tests: there is an existing test on [ProtectedBranches](https://github.com/ubisoft/NGitLab/blob/main/NGitLab.Tests/ProtectedBranchTests.cs).
> 
> You could likely make a little one to check the updates. If you don't have docker locally, we can still validate the test on the CI

@Toa741 I have updated the existing protected branch test with a test on the update, but I'm not sure my changes are correct. Locally, using Docker, the original tests fails with a 404 on the ProtectBranch call. I don't know why. Any ideas?

Toa741:
I do see the not found too, that's strange. I'll try to see what's going on

Toa741:
> > About the tests: there is an existing test on [ProtectedBranches](https://github.com/ubisoft/NGitLab/blob/main/NGitLab.Tests/ProtectedBranchTests.cs).
> > You could likely make a little one to check the updates. If you don't have docker locally, we can still validate the test on the CI
> 
> @Toa741 I have updated the existing protected branch test with a test on the update, but I'm not sure my changes are correct. Locally, using Docker, the original tests fails with a 404 on the ProtectBranch call. I don't know why. Any ideas?

Ok so, it seems related to the new property you added in `AccessLevelInfo`. Since it's used in the `BranchProtect` class, by default on creation it will transmit the default value (0) for Id,  UserId & GroupId. I guess it doesn't like that.

Making these new properties as nullable would likely fix the problem.

Although, not really ideal. Not sure there is another choise though, Ideally the BranchProtect would have used some sort of `AccessLevelInfoCreate`, but that doesn't exists :/

You can try with that, you should be able to push the test further.

I'll add a note to have some `AccessLevelInfoCreate` or something equivalent in a next major version

JesseKlaasse:
> > > About the tests: there is an existing test on [ProtectedBranches](https://github.com/ubisoft/NGitLab/blob/main/NGitLab.Tests/ProtectedBranchTests.cs).
> > > You could likely make a little one to check the updates. If you don't have docker locally, we can still validate the test on the CI
> > 
> > 
> > @Toa741 I have updated the existing protected branch test with a test on the update, but I'm not sure my changes are correct. Locally, using Docker, the original tests fails with a 404 on the ProtectBranch call. I don't know why. Any ideas?
> 
> Ok so, it seems related to the new property you added in `AccessLevelInfo`. Since it's used in the `BranchProtect` class, by default on creation it will transmit the default value (0) for Id, UserId & GroupId. I guess it doesn't like that.
> 
> Making these new properties as nullable would likely fix the problem.
> 
> Although, not really ideal. Not sure there is another choise though, Ideally the BranchProtect would have used some sort of `AccessLevelInfoCreate`, but that doesn't exists :/
> 
> You can try with that, you should be able to push the test further.
> 
> I'll add a note to have some `AccessLevelInfoCreate` or something equivalent in a next major version

I have done this, and updated the tests. They now succeed.

PR #872, opened 2025-03-12
Add property ci_default_git_depth to both Project.cs and ProjectUpdate.cs

matt-pan:
This property can be set via PUT update on projects and is part of the Project response


matt-pan:
Oh, I didn't know this is tested against the public API
Should be fine though, as this parameter is in https://gitlab.com/gitlab-org/gitlab/-/blob/master/doc/api/openapi/openapi_v2.yaml on the GET /api/v4/projects/{id} as well as the PUT /api/v4/projects/{id}

NGitLab/Models/ProjectUpdate.cs
```diff
@@ -111,4 +111,7 @@ public sealed class ProjectUpdate
 
     [JsonPropertyName("topics")]
     public List<string> Topics { get; set; }
+
+    [JsonPropertyName("ci_default_git_depth")]
+    public bool? CiDefaultGitDepth { get; set; }
```
louis-z: This property should be an `int?` instead. Refer to [the GitLab doc](https://docs.gitlab.com/api/projects/#edit-a-project).

NGitLab/Models/Project.cs
```diff
@@ -224,4 +224,7 @@ public class Project
 
     [JsonPropertyName("mr_default_target_self")]
     public bool? MrDefaultTargetSelf { get; set; }
+
+    [JsonPropertyName("ci_default_git_depth")]
+    public bool? CiDefaultGitDepth { get; set; }
```
louis-z: This property should be an `int?` instead. Refer to [the GitLab doc](https://docs.gitlab.com/api/projects/#get-a-single-project).

louis-z (changes requested):
Thanks for your contribution, @matt-pan.

In order to get this PR merged quickly,
- Address the inline comments, and
- Update the declared public API. This can easily be done via VS2022's "Quick Actions and Refactorings" _(when asked "Fix all occurrences in:", make sure to select "Solution")_:
image

PR #874, opened 2025-03-13
Add parameters to Repository compare

Codex04:


NGitLab/Models/CompareQuery.cs
```diff
@@ -15,6 +17,22 @@ public class CompareQuery
     /// </summary>
     public string Target { get; set; }
 
+    /// <summary>
+    /// Comparison method: true for direct comparison between from and to (from..to), false to compare using merge base (from…to)’. Default is false.
+    /// </summary>
+    public bool Straight { get; set; }
```
louis-z: Since [`straight` is an optional query parameter](https://docs.gitlab.com/api/repositories/#compare-branches-tags-or-commits) (and even though its default implicit value is `false`), I would make it nullable. Same thing for `unidiff`.

NGitLab/Models/CompareQuery.cs
```diff
@@ -15,6 +17,22 @@ public class CompareQuery
...
+
+    /// <summary>
+    /// Present diffs in the unified diff format https://www.gnu.org/software/diffutils/manual/html_node/Detailed-Unified.html. Default is false. Introduced in GitLab 16.5.
+    /// </summary>
+    public bool Unidiff { get; set; }
+
+    /// <summary>
+    /// The ID to compare from.
+    /// </summary>
+    [JsonPropertyName("from_project_id")]
```
louis-z: Unnecessary, as it's not serialized to/deserialized from JSON.

NGitLab/Impl/RepositoryClient.cs
```diff
@@ -109,7 +109,11 @@ public IEnumerable<Commit> GetCommits(GetCommitsRequest request)
 
     public CompareResults Compare(CompareQuery query)
     {
-        return _api.Get().To<CompareResults>(_repoPath + $@"/compare?from={query.Source}&to={query.Target}");
+        var url = _repoPath + $"/compare?from={query.Source}&to={query.Target}";
```
louis-z: `CompareAsync` also needs to be adjusted. You should probably move to a new centralized method the construction of the URL, to avoid repeating code.

louis-z (changes requested):

PR #897, opened 2025-05-01
Add support for the "Get raw file" API

ap0llo:
Add support for downloading the content's of a file via GitLabs "Get raw file" API. 

In contrast to the existing "Get" method that gets a file's content and its metadata, the "GetRaw" method gives access to the file stream. 
This avoids having to loads the contents of the entire file into memory. 
Additionally, this API allows getting the content of the file if the file was commited via Git LFS ("Get()" will just return the pointer to the LFS file in this case)

See Also [Get raw file from repository (GitLab Docs)](https://docs.gitlab.com/api/repository_files/#get-raw-file-from-repository)

### DIFF
--- a/NGitLab.Mock/Repository.cs
+++ b/NGitLab.Mock/Repository.cs
@@ -538,6 +538,32 @@ public sealed class Repository : GitLabObject, IDisposable
         };
     }
 
+    public void GetRawFile(string filePath, Action<Stream> parser, GetRawFileRequest request)
+    {
+        var repo = GetGitRepository();
+        try
+        {
+            var @ref = request?.Ref ?? repo.Head.FriendlyName;
+            Commands.Checkout(repo, @ref);
+        }
+        catch (LibGit2Sharp.NotFoundException)
+        {
+            throw GitLabException.NotFound("File not found");
+        }
+
+        var fileCompletePath = Path.Combine(FullPath, filePath);
+
+        if (!System.IO.File.Exists(fileCompletePath))
+        {
+            throw GitLabException.NotFound("File not found");
+        }
+
+        using (var fileStream = System.IO.File.OpenRead(fileCompletePath))
+        {
+            parser(fileStream);
+        }
+    }
+
     internal IEnumerable<Tree> GetTree() => GetTree(new());
 
     internal IEnumerable<Tree> GetTree(RepositoryGetTreeOptions repositoryGetTreeOptions)
--- a/NGitLab.Mock/Repository.cs
+++ b/NGitLab.Mock/Repository.cs
@@ -548,7 +548,7 @@ public sealed class Repository : GitLabObject, IDisposable
         }
         catch (LibGit2Sharp.NotFoundException)
         {
-            throw GitLabException.NotFound("File not found");
+            throw GitLabException.NotFound("Revision not found");
         }
 
         var fileCompletePath = Path.Combine(FullPath, filePath);
--- a/NGitLab.Mock/Repository.cs
+++ b/NGitLab.Mock/Repository.cs
@@ -5,6 +5,7 @@ using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Threading;
+using System.Threading.Tasks;
 using LibGit2Sharp;
 using NGitLab.Mock.Clients;
 using NGitLab.Models;
@@ -538,7 +539,7 @@ public sealed class Repository : GitLabObject, IDisposable
         };
     }
 
-    public void GetRawFile(string filePath, Action<Stream> parser, GetRawFileRequest request)
+    public async Task GetRawFileAsync(string filePath, Func<Stream, Task> parser, GetRawFileRequest request)
     {
         var repo = GetGitRepository();
         try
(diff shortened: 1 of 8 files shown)


NGitLab/GetRawFileRequest.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace NGitLab;
+
+public class GetRawFileRequest
+{
+    /// <summary>
+    /// The revision of the file to get (name of the branch/tag or commit)
```
louis-z:
```suggestion
    /// The branch name, tag or commit to get the file from. Default is the HEAD of the project.
```

NGitLab/GetRawFileRequest.cs
```diff
@@ -0,0 +1,14 @@
...
+    /// </summary>
+    public string Ref { get; set; } = null;
```
louis-z: - Don't initialize to default value
- When possible, declare `init` accessor rather than `set`

```suggestion
    public string Ref { get; init; }
```

NGitLab/Impl/FilesClient.cs
```diff
@@ -57,6 +58,23 @@ public Task<FileData> GetAsync(string filePath, string @ref, CancellationToken c
         return _api.Get().ToAsync<FileData>(_repoPath + $"/files/{EncodeFilePath(filePath)}?ref={Uri.EscapeDataString(@ref)}", cancellationToken);
     }
 
+    public void GetRaw(string filePath, Action<Stream> parser, GetRawFileRequest request = null)
+    {
+        var url = _repoPath + $"/files/{EncodeFilePath(filePath)}/raw/";
```
louis-z:
```suggestion
        var url = _repoPath + $"/files/{EncodeFilePath(filePath)}/raw";
```

NGitLab/Impl/FilesClient.cs
```diff
@@ -57,6 +58,23 @@ public Task<FileData> GetAsync(string filePath, string @ref, CancellationToken c
...
+
+        if (!String.IsNullOrWhiteSpace(request?.Ref))
```
louis-z:
```suggestion
        if (!string.IsNullOrWhiteSpace(request?.Ref))
```

NGitLab/Impl/FilesClient.cs
```diff
@@ -57,6 +58,23 @@ public Task<FileData> GetAsync(string filePath, string @ref, CancellationToken c
...
+        {
+            url = Utils.AddParameter(url, "ref", Uri.EscapeDataString(request?.Ref));
```
louis-z: We've already made sure that `request` isn't null:

```suggestion
            url = Utils.AddParameter(url, "ref", Uri.EscapeDataString(request.Ref));
```

NGitLab/Impl/FilesClient.cs
```diff
@@ -57,6 +58,23 @@ public Task<FileData> GetAsync(string filePath, string @ref, CancellationToken c
...
+        }
+
+        if (request is not null && request.Lfs.HasValue)
+        {
+            url = Utils.AddParameter(url, "lfs", request?.Lfs.Value);
```
louis-z:
```suggestion
            url = Utils.AddParameter(url, "lfs", request.Lfs.Value);
```

NGitLab.Mock/Repository.cs
```diff
@@ -538,6 +538,32 @@ public FileData GetFile(string filePath, string @ref)
...
+    {
+        var repo = GetGitRepository();
+        try
+        {
+            var @ref = request?.Ref ?? repo.Head.FriendlyName;
+            Commands.Checkout(repo, @ref);
+        }
+        catch (LibGit2Sharp.NotFoundException)
+        {
+            throw GitLabException.NotFound("File not found");
```
louis-z: It's not the file that wasn't found, but rather the `@ref`.

NGitLab.Mock/Repository.cs
```diff
@@ -538,6 +538,32 @@ public FileData GetFile(string filePath, string @ref)
...
+        }
+
+        var fileCompletePath = Path.Combine(FullPath, filePath);
+
+        if (!System.IO.File.Exists(fileCompletePath))
```
louis-z, line 557: Please simplify:

```suggestion
        if (!File.Exists(fileCompletePath))
```

NGitLab/IFilesClient.cs
```diff
@@ -22,6 +24,8 @@ public interface IFilesClient
 
     Task<FileData> GetAsync(string filePath, string @ref, CancellationToken cancellationToken = default);
 
+    void GetRaw(string filePath, Action<Stream> parser, GetRawFileRequest request = null);
```
louis-z: Can you make the method async and rename it with the Async suffix?

louis-z (changes requested):

NGitLab.Mock/Repository.cs
ap0llo: Unfortunately, the fully-qualified name is required here as it conflicts with type `NGitLab.Mock.File`

ap0llo:
@louis-z Thanks for your review, I've incorporated your feedback.

NGitLab.Mock/Repository.cs
louis-z: Fair enough.

PR #899, opened 2025-05-02
fix!: Job Artifact-related API inconsistencies

louis-z:
When retrieving a job's data, GitLab can return both `artifacts_file` and `artifacts`

```json
{
  "artifacts_file": {
    "filename": "artifacts.zip",
    "size": 1000
  },
  "artifacts": [
    {
      "file_type": "archive",
      "size": 1000,
      "filename": "artifacts.zip",
      "file_format": "zip"
    },
    {
      "file_type": "metadata",
      "size": 186,
      "filename": "metadata.gz",
      "file_format": "gzip"
    },
    {
      "file_type": "trace",
      "size": 1500,
      "filename": "job.log",
      "file_format": "raw"
    },
    {
      "file_type": "junit",
      "size": 750,
      "filename": "junit.xml.gz",
      "file_format": "gzip"
    }
  ]
}
```

- Rename `NGitLab.Models.Artifacts` to `NGitLab.Models.Artifacts.ArtifactsFile`
- Add the `NGitLab.Models.Artifacts` array
- Adapt Mocks and Tests accordingly

### DIFF
--- a/NGitLab.Mock/Clients/JobClient.cs
+++ b/NGitLab.Mock/Clients/JobClient.cs
@@ -127,7 +127,8 @@ internal sealed class JobClient : ClientBase, IJobClient
                     job.Status = JobStatus.Canceled;
                     break;
                 case JobAction.Erase:
-                    job.Artifacts = null;
+                    job.ArtifactsFile = null;
+                    job.Artifacts.Clear();
                     job.Trace = null;
                     break;
                 case JobAction.Play:
--- a/NGitLab.Mock/Job.cs
+++ b/NGitLab.Mock/Job.cs
@@ -1,11 +1,17 @@
 ﻿using System;
 using System.Globalization;
+using System.Linq;
 using NGitLab.Models;
 
 namespace NGitLab.Mock;
 
 public sealed class Job : GitLabObject
 {
+    public Job()
+    {
+        Artifacts = new JobArtifactCollection(this);
+    }
+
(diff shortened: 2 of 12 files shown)


theknut:
Does this fix the problem that artifacts should be gone due to policy but they are still the latest for this branch and therefore kept? Previously this resulted in `artifacts_file` being `null`. The workaround was to use the api to retrieve the `artifacts` property and check if any items contains `archive`. If this PR fixes this, this would mean that a check for `job.Artifacts` considers both cases now?

louis-z:
> Does this fix the problem that artifacts should be gone due to policy but they are still the latest for this branch and therefore kept? Previously this resulted in `artifacts_file` being `null`. The workaround was to use the api to retrieve the `artifacts` property and check if any items contains `archive`. If this PR fixes this, this would mean that a check for `job.Artifacts` considers both cases now?

This PR

- Corrects mappings as follows:

  | GitLab API JSON field | NGitLab property | comments |
  | --- | --- | --- |
  | "artifacts" | `Job.Artifacts` | JSON field was previously not mapped to any NGitLab property |
  | "artifacts_file" | `Job.ArtifactsFile` | JSON field used to wrongly map to `Job.Artifacts` NGitLab property |

- Does not add any other specific logic.

theknut:
Okay, I see. The mapping is now correct and now I can access both properties without any workarounds. Thanks!

PR #902, opened 2025-05-15
Make NGitLab.Mock.Clients.ClientContext usable in an async context

ap0llo:
In ClientContext, use SemaphoreSlim instead of Monitor to synchronize operations.
This allows using the ClientContext in an async method.

The Monitor-based implementation throws an SynchronizationLockException when attempting to release the lock since releasing the lock might happen on a different thread than the thread that entered the monitor.

As far as I can tell, NGitLab.Mock.Clients.FileClient.GetRawAsync() is the first occurrence where an async operation happens within an operation scope, so this was not an issue before this method was added.

### DIFF
--- a/NGitLab.Mock/Clients/ClientContext.cs
+++ b/NGitLab.Mock/Clients/ClientContext.cs
@@ -1,11 +1,12 @@
 ﻿using System;
 using System.Threading;
+using System.Threading.Tasks;
 
 namespace NGitLab.Mock.Clients;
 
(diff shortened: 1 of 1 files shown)


louis-z:
Thanks @ap0llo for your contribution. I created PR #910, based on yours but with slight changes so that tests wouldn't hang. That PR was merged last week, rendering this one unnecessary.

ap0llo:
Thank you!

PR #908, opened 2025-06-10
feat: add job runner manager property

theknut:
Currently only `runner` information is set to `Job.Runner`. This PR also processes `runner_manager` and introduces a new property `Job.RunnerManager`.

### DIFF
--- a/NGitLab/Models/Job.cs
+++ b/NGitLab/Models/Job.cs
@@ -10,6 +10,9 @@ public class Job : JobBasic
     [JsonPropertyName("runner")]
     public JobRunner Runner { get; set; }
 
+    [JsonPropertyName("runner_manager")]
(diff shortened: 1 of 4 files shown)


NGitLab/Models/Job.cs
```diff
@@ -10,6 +11,9 @@ public class Job : JobBasic
     [JsonPropertyName("runner")]
     public JobRunner Runner { get; set; }
 
+    [JsonPropertyName("runner_manager")]
+    public JobRunner RunnerManager { get; set; }
```
Toa741: Is it intended to have the type `JobRunnerManager`?

theknut: Yes, sorry. Thats the problem if you just edit the file on GitHub and never open it in a code editor...

PR #917, opened 2025-06-24
Implement Get merge request diffs in MergeRequestClient & Merge base from refs

Codex04:
Implementations:
- Get merge request diff (https://docs.gitlab.com/api/merge_requests/#list-merge-request-diffs)
- Get versioned merge request diff (https://docs.gitlab.com/api/merge_requests/#get-a-single-merge-request-diff-version)
- Get merge base commit from refs (https://docs.gitlab.com/api/repositories/#get-merge-base)

### DIFF
--- a/NGitLab/Models/VersionedDiffResult.cs
+++ b/NGitLab/Models/VersionedDiffResult.cs
@@ -1,0 +1,40 @@
+﻿using System;
+using System.Text.Json.Serialization;
+
+namespace NGitLab.Models;
+
+public class VersionedDiffResult
+{
+    [JsonPropertyName("id")]
+    public long Id { get; set; }
+
+    [JsonPropertyName("base_commit_sha")]
+    public long BaseCommitSha { get; set; }
+
+    [JsonPropertyName("commits")]
+    public Commit[] Commits { get; set; }
+
+    [JsonPropertyName("created_at")]
+    public DateTime CreatedAt { get; set; }
+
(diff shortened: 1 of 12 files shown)


NGitLab/Models/VersionedDiffResult.cs
```diff
@@ -0,0 +1,40 @@
...
+
+namespace NGitLab.Models;
+
+public class VersionedDiffResult
+{
+    [JsonPropertyName("id")]
+    public long Id { get; set; }
+
+    [JsonPropertyName("base_commit_sha")]
+    public long BaseCommitSha { get; set; }
```
louis-z: According to the doc, this should be a `string`. However, since it represents a SHA, it could also be of type `Sha1`.

NGitLab/Models/VersionedDiffResult.cs
```diff
@@ -0,0 +1,40 @@
...
+
+    [JsonPropertyName("commits")]
+    public Commit[] Commits { get; set; }
+
+    [JsonPropertyName("created_at")]
+    public DateTime CreatedAt { get; set; }
+
+    [JsonPropertyName("diffs")]
+    public Diff[] Diff { get; set; }
```
louis-z:
```suggestion
    public Diff[] Diffs { get; set; }
```

louis-z (changes requested):

PR #921, opened 2025-06-26
Added method to edit environment description

fakefeik:
Added method `EditDescription(long environmentId, string description)` to IEnvironmentsClient


fakefeik:
realized I should use POST with body (description can be multiline)

PR #982, opened 2025-10-25
Add Package Tags

pos777:
[Package authoring best practices / Tags](https://learn.microsoft.com/en-us/nuget/create-packages/package-authoring-best-practices#tags)

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -18,6 +18,7 @@
     <!-- Package info -->
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <PackageReadmeFile>README.md</PackageReadmeFile>
+    <PackageTags>GitLab REST API Client</PackageTags>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


Directory.Build.props
```diff
@@ -18,6 +18,7 @@
     <!-- Package info -->
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <PackageReadmeFile>README.md</PackageReadmeFile>
+    <PackageTags>GitLab REST API Client</PackageTags>
```
Toa741:
```suggestion
    <PackageTags>GitLab REST API Client;GitLab;GitLab API;GitLab client</PackageTags>
```

Toa741:
@pos777 , I went over this example: https://learn.microsoft.com/en-us/nuget/create-packages/creating-a-package-msbuild#add-an-optional-description-field

But seems it's not a "good" example.
More info there about it: https://github.com/dotnet/docs/issues/13553#issuecomment-518444966

Will just keep the original suggestion you did.

Thanks :),

PR #1000, opened 2025-11-11
feat: add sha & deployable to deployment

mizzm001:
adds sha & deployable to deployment


mizzm001:
@louis-z I have pushed necessary changes for nuget build

mizzm001:
@louis-z I have fixed the necessary line spacings

PR #1047, opened 2026-02-03
feat: Implement TagClient.GetAsync in NGitLab.Client.Mock

mlemieuxlafontaine-ubi:
Well to be exact, the method was already implemented, but we could not pass in a query without the method crashing. Implemented the behavior as closely as possible to the GitLab's api sorting and filtering

### DIFF
--- a/NGitLab.Mock/Clients/TagClient.cs
+++ b/NGitLab.Mock/Clients/TagClient.cs
@@ -6,7 +6,6 @@ using System.Threading;
 using System.Threading.Tasks;
 using NGitLab.Mock.Internals;
 using NGitLab.Models;
-using NuGet.Versioning;
 using Commit = LibGit2Sharp.Commit;
 
 namespace NGitLab.Mock.Clients;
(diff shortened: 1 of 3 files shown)


NGitLab.Mock/Clients/TagClient.cs
```diff
@@ -85,17 +86,36 @@ public Tag ToTagClient(LibGit2Sharp.Tag tag)
...
+            {
+                "name" => tags.OrderBy(t => t.FriendlyName, StringComparer.Ordinal),
+                "updated" or null => tags,
+                "version" => tags.OrderBy(t => SemanticVersion.TryParse(t.FriendlyName, out var s) ? s : null),
+                _ => throw new NotImplementedException(),
+            };
+
+            return direction switch
+            {
+                null or "desc" => tags.Reverse(),
```
Toa741: -> null never happens
-> NotImplementedException : NotSupported("Direction {direction} is not supported")

NGitLab.Mock/Clients/TagClient.cs
```diff
@@ -85,17 +86,36 @@ public Tag ToTagClient(LibGit2Sharp.Tag tag)
...
+        {
+            if (string.IsNullOrEmpty(direction))
+                direction = "desc";
+
+            // LibGitSharp does not really expose tag creation time, so hard to sort using that annotation,
+            // we'll skip sorting in that case (which should be functionnaly similar to "name" sorting)
+            tags = orderBy switch
+            {
+                "name" => tags.OrderBy(t => t.FriendlyName, StringComparer.Ordinal),
+                "updated" or null => tags,
```
Toa741: -> null: no orderby
-> updated: throw

PR #1048, opened 2026-02-04
fix!: SYSLIB0014 warning about use of WebRequest-related types

louis-z:
Replace use of obsolete WebRequest-related types with that of HttpClient-related ones.

This was done in several steps, ultimately squashed into a single commit:

- Claude did the heavy lifting
- Manual tweaks were made afterwards, to make code changes a little cleaner

This introduces Breaking Changes:
- ⚠ Users can no longer override `GetResponse()`, `GetResponseAsync()`, or `GetRequestStream()` methods
- ✅ Migration path: Use `IHttpMessageHandler` interface instead


.claude/settings.local.json
```diff
@@ -0,0 +1,10 @@
+{
```
msrouchou: *.local.json must be gitignored and not exposed to git ^^

msrouchou:
- Are there any changes as to how the client library will be used in consuming apps ? 
- In case the breaking changes are confusing, can we create/generate a quick migration guide  (from older to newer version) ?

PR #1108, opened 2026-06-22
Add container registry api support

msrouchou:
Adds `IContainerRegistryClient` with methods to list repositories, list tags, delete a tag, and delete a repository. Includes mock implementation for tests.


NGitLab.Mock/Project.cs
```diff
@@ -163,6 +164,8 @@ public string[] Tags
 
     public ProtectedBranchCollection ProtectedBranches { get; }
 
+    public ContainerRepositoryCollection ContainerRepositories { get; }
```
Toa741, line 167: Can you add the possibility to set it in the GitLabHelpers (to construct GitLabConfig), thus, to be able to use the mocked client.

msrouchou: fixed ^^

PR #1110, opened 2026-06-25
Add permanently_remove support to project delete

msrouchou:
Exposes the `permanently_remove` and `full_path` query parameters on `IProjectClient.DeleteAsync,` matching the GitLab API behavior where v18+ soft-deletes by default and requires an explicit flag to bypass the grace period.

https://docs.gitlab.com/api/projects/#delete-a-project

### DIFF
--- a/NGitLab.Mock.Tests/ProjectsMockTests.cs
+++ b/NGitLab.Mock.Tests/ProjectsMockTests.cs
@@ -442,9 +442,9 @@ public class ProjectsMockTests
     }
 
     [Test]
-    public async Task DeleteAsync_WhenProjectExists_ItIsMarkedForDeletion()
+    public async Task DeleteAsync_WhenProjectExists_MarksProjectForDeletion()
     {
-        var projectFullPath = $"Test/{nameof(DeleteAsync_WhenProjectExists_ItIsMarkedForDeletion)}";
+        var projectFullPath = $"Test/{nameof(DeleteAsync_WhenProjectExists_MarksProjectForDeletion)}";
         using var server = new GitLabConfig()
             .WithUser("Test", isDefault: true)
             .WithProjectOfFullPath(projectFullPath)
@@ -494,6 +494,7 @@ public class ProjectsMockTests
         // Assert
         var markedProject = await projectClient.GetAsync(project.Id);
         Assert.That(markedProject.MarkedForDeletionOn, Is.Not.Null);
+        // The mock sets MarkedForDeletionOn to UtcNow at the time of soft-deletion.
         Assert.That(markedProject.MarkedForDeletionOn!.Value.Date, Is.EqualTo(DateTime.UtcNow.Date));
     }
 
@@ -509,7 +510,7 @@ public class ProjectsMockTests
         var projectClient = server.CreateClient().Projects;
         var project = await projectClient.GetAsync(projectFullPath);
 
-        // First call: soft-mark
+        // GitLab requires the project to be soft-deleted before permanently_remove can be used.
         await projectClient.DeleteAsync(project.Id);
 
         // Second call: permanently remove
@@ -535,7 +536,7 @@ public class ProjectsMockTests
         var projectClient = server.CreateClient().Projects;
         var project = await projectClient.GetAsync(projectFullPath);
 
-        // Soft-mark first
+        // GitLab requires the project to be soft-deleted before permanently_remove can be used.
         await projectClient.DeleteAsync(project.Id);
 
         // Act: wrong full_path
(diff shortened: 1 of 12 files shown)


NGitLab.Mock.Tests/ProjectsMockTests.cs
```diff
@@ -442,9 +442,9 @@ public void UpdateAsync_WhenProjectNotFound_ItThrows()
     }
 
     [Test]
-    public async Task DeleteAsync_WhenProjectExists_ItIsDeleted()
+    public async Task DeleteAsync_WhenProjectExists_ItIsMarkedForDeletion()
```
louis-z: For better consistency with the additional test that follows this one:

```suggestion
    public async Task DeleteAsync_WhenProjectExists_MarksProjectForDeletion()
```

NGitLab.Mock.Tests/ProjectsMockTests.cs
```diff
@@ -475,6 +476,80 @@ public void DeleteAsync_WhenProjectNotFound_ItThrows()
...
+        var projectClient = server.CreateClient().Projects;
+        var project = await projectClient.GetAsync(projectFullPath);
+
+        // Act
+        await projectClient.DeleteAsync(project.Id);
+
+        // Assert
+        var markedProject = await projectClient.GetAsync(project.Id);
+        Assert.That(markedProject.MarkedForDeletionOn, Is.Not.Null);
+        Assert.That(markedProject.MarkedForDeletionOn!.Value.Date, Is.EqualTo(DateTime.UtcNow.Date));
```
louis-z, line 498: How come deletion is scheduled for today? I'm assuming this is due to a server setting, but can you provide a comment that explains it?

NGitLab.Mock.Tests/ProjectsMockTests.cs
```diff
@@ -475,6 +476,80 @@ public void DeleteAsync_WhenProjectNotFound_ItThrows()
...
+        using var server = new GitLabConfig()
+            .WithUser("Test", isDefault: true)
+            .WithProjectOfFullPath(projectFullPath)
+            .BuildServer();
+
+        var projectClient = server.CreateClient().Projects;
+        var project = await projectClient.GetAsync(projectFullPath);
+
+        // First call: soft-mark
+        await projectClient.DeleteAsync(project.Id);
```
louis-z, line 514: Is this call really needed, especially since we don't assert anything after it? Besides, the soft delete path is already tested elsewhere.

NGitLab.Mock.Tests/ProjectsMockTests.cs
```diff
@@ -475,6 +476,80 @@ public void DeleteAsync_WhenProjectNotFound_ItThrows()
...
+        using var server = new GitLabConfig()
+            .WithUser("Test", isDefault: true)
+            .WithProjectOfFullPath(projectFullPath)
+            .BuildServer();
+
+        var projectClient = server.CreateClient().Projects;
+        var project = await projectClient.GetAsync(projectFullPath);
+
+        // Soft-mark first
+        await projectClient.DeleteAsync(project.Id);
```
louis-z, line 540: Again, why is this here?

NGitLab/Impl/ProjectClient.cs
```diff
@@ -39,8 +40,26 @@ public Task<Project> CreateAsync(ProjectCreate project, CancellationToken cancel
...
+    public Task DeleteAsync(ProjectId projectId, ProjectDelete options, CancellationToken cancellationToken = default)
+    {
+        var url = $"{Project.Url}/{projectId.ValueAsUriParameter()}";
+        if (options is not null)
+        {
+            url = Utils.AddParameter(url, "permanently_remove", options.PermanentlyRemove);
+            if (options.FullPath is not null)
+            {
+                var @operator = url.Contains('?') ? "&" : "?";
+                url += $"{@operator}full_path={options.FullPath}";
```
louis-z:
```suggestion
                url = Utils.AddParameter(url, "full_path", options.FullPath);
```

louis-z (changes requested):

NGitLab.Mock.Tests/ProjectsMockTests.cs
msrouchou: Fixed.

NGitLab.Mock.Tests/ProjectsMockTests.cs
msrouchou: The mock sets `MarkedForDeletionOn` to `DateTime.UtcNow` at soft-delete time. Added a comment to clarify.

NGitLab.Mock.Tests/ProjectsMockTests.cs
msrouchou: It's a required setup step. GitLab only allows `permanently_remove` on a project that's already soft-deleted. Without this first call the second would return 400. Clarified the comment.

NGitLab.Mock.Tests/ProjectsMockTests.cs
msrouchou: Same as above, it's the required soft-delete step before `permanently_remove`.

NGitLab/Impl/ProjectClient.cs
msrouchou: Fixed.

PR #1140, opened 2026-09-10
feat: implement getversionsasync in mock

FBSALOMON:
Implement GetVersionsAsync in MergeRequestClient to be able to add comments and threads into reviews.
Add Comment, Discussion, Versions mock tests
Simplify CreateMergeRequestAsync:  
1. Waiting for the Merge Requests to be 'mergable' this avoid return them while the DIFF Versions are still empty this was causing  a lot of retries.
2. Remove the Push into the main branch and Creating a Branch at the same time, this would sometimes return a CONFLICT when it should not.


NGitLab.Mock/Clients/MergeRequestClient.cs
```diff
@@ -728,7 +728,15 @@ public Models.MergeRequest Update(long mergeRequestIid, MergeRequestUpdate merge
 
     public GitLabCollectionResponse<MergeRequestVersion> GetVersionsAsync(long mergeRequestIid)
     {
-        throw new NotImplementedException();
+        AssertProjectId();
+
+        using (Context.BeginOperationScope())
+        {
+            var mergeRequest = GetMergeRequest(_projectId.GetValueOrDefault(), mergeRequestIid);
```
msrouchou, line 735: can this call already return the list ordered as we want, instead of a second call to sort ?

NGitLab.Mock/MergeRequest.cs
```diff
@@ -400,5 +410,16 @@ private void RefreshInternalState()
         {
             _hasConflicts = true;
         }
+
+        _versions.Add(new Models.MergeRequestVersion
```
msrouchou, line 414: to verify if Add is ok or if we need to make _versions a Set

NGitLab.Tests/Docker/GitLabTestContext.cs
```diff
@@ -493,14 +487,17 @@ public static async Task<T> RetryUntilAsync<T>(Func<Task<T>> action, Func<T, Tas
         }
     }
 
-    public static async Task<T> RetryUntilAsync<T>(Func<T> action, Func<T, bool> predicate, CancellationToken cancellationToken)
+    // INVESTIGATION SCAFFOLDING: `operationName` (optional, so existing call sites are unaffected) identifies
+    // which of the ~30 call sites is retrying, since the shared test output otherwise interleaves indistinguishable
+    // "RetryUntilAsync {n}..." lines from concurrent/sequential polls (e.g. branch-readiness vs. mergeable-status).
```
msrouchou: comment seems relevant to a coding session and not interesting for the end user ? wither we rework the comment or remove it

NGitLab.Mock/Clients/MergeRequestClient.cs
msrouchou: discussed privately, no GetMergeRequest override to give more options

NGitLab.Mock/MergeRequest.cs
msrouchou: validated with FBSALOMON

PR #1149, opened 2026-09-16
Docker.DotNet → Testcontainers migration

louis-z:
- Replace the `Docker.DotNet` package (unmaintained) with `Docker.DotNet.Enhanced` (its actively maintained fork, now owned by the Testcontainers org).
- Add the `Testcontainers` package.
- `GitLabDockerContainer.SpawnLocalDockerContainerAsync` (used for local test runs) now creates/starts the GitLab container via Testcontainers' `ContainerBuilder`, instead of ~90 lines of manual container lookup/create/start/health-polling code.
- Credential generation (`RunGitLabRailsRunnerAsync`) now runs the `gitlab-rails runner` command differently depending on how the container came to exist:
  - **Local runs**: execs directly into the container object Testcontainers already holds a reference to.
  - **CI runs**: GitLab is started by GitHub Actions as a service container, not by our code, so we still use a raw Docker Engine API client (via `Docker.DotNet.Enhanced`) to find that container and exec into it.

## Rationale

- `Docker.DotNet` has had no releases or fixes in a long time; staying on it was a growing maintenance risk.
- `Testcontainers` is the purpose-built library for "spin up a container for tests, wait until it's ready, tear it down"; it replaces our hand-rolled orchestration/health-check code instead of just swapping the underlying HTTP client.
- Testcontainers can't attach to a container it didn't create itself, so the CI path (which execs into a pre-existing GitHub Actions service container) still needs a low-level Docker API client. `Docker.DotNet.Enhanced` fills that one remaining need, and since Testcontainers depends on it internally too, there's no duplicate/conflicting dependency.

## Trade-off to know about

- The old code auto-detected when the local GitLab image version changed and removed the stale container before creating a new one.
- The new code uses Testcontainers' `WithReuse(true)` to keep the container running across local test runs (GitLab takes minutes to boot), but reuse matching is based on name + config.
- **If you bump `LocalGitLabDockerVersion`, you now need to manually remove the old container first:**
  ```
  docker rm -f NGitLabClientTests
  ```


NGitLab.Tests/Docker/GitLabDockerContainer.cs
msrouchou: `ContainerBuilder` does not accept the image name as a positional constructor argument. The public Testcontainers.NET 4.x API is fluent:

```csharp
_localContainer = new ContainerBuilder()
    .WithImage(ImageName + ":" + LocalGitLabDockerVersion)
    ...
    .Build();
```

If the overload `new ContainerBuilder(string imageName)` does not exist in 4.15.0 this will fail to compile.

msrouchou: `_localContainer` (`IContainer`) is `IAsyncDisposable` but `DisposeAsync` is never called on it. The underlying `DockerClient`, factory, and HTTP resources leak for every local test run. (`WithReuse` keeps the *container process* alive across runs, but the .NET wrapper object is a separate concern.)

Implement `IAsyncDisposable` on `GitLabDockerContainer` and call `await _localContainer.DisposeAsync()` in the override.

msrouchou: The old `SpawnDockerContainerAsync` called `ValidateDockerIsEnabled` before touching the Docker API, which printed a clear message ("Cannot connect to Docker service. Make sure it's running on your machine.") and then skipped the test via `Assert.Ignore`. `SpawnLocalDockerContainerAsync` omits this call, so if Docker is not running locally the error surfaces as a raw Testcontainers startup exception with no guidance toward the root cause.

Consider calling `ValidateCiDockerIsEnabled` (or a local-only variant) at the top of `SpawnLocalDockerContainerAsync` to preserve the diagnostic.

msrouchou: `AdminUserName` is interpolated verbatim into the Ruby heredoc (`User.find_by_username!('{AdminUserName}')`). It is currently the constant `"root"` so it's safe, but there is no sanitisation. If it ever contains a single quote or Ruby metacharacter the generated script will be syntactically broken or behave unexpectedly.

Pass the value as an environment variable on the exec call and read it with `ENV.fetch('ADMIN_USERNAME')` inside Ruby, or apply single-quote escaping before interpolating.

msrouchou: `RunGitLabRailsRunnerAsync(DockerClient client, string script)`: the `client` parameter is `null` on the local path and a live `DockerClient` on the CI path. The `else` branch dereferences `client` unconditionally (`ResolveCiGitLabContainerIdAsync(client)`, then `client.Exec.*`). The invariant `_localContainer != null ↔ client == null` is upheld at every current call site, but there is no assertion, annotation, or type to enforce it — a future call with both `_localContainer == null` and `client == null` would produce a `NullReferenceException`.

Consider splitting into two private methods (one per path) so each takes only the arguments it needs and the nullable parameter disappears entirely.

NGitLab.Tests/Docker/GitLabDockerContainer.cs
```diff
@@ -289,43 +205,45 @@ private async Task GenerateCredentialsAsync()
         Console.WriteLine("Requesting credentials from GitLab instance");
 
         var credentials = new GitLabCredential();
```
msrouchou, line 207: `All = true` returns stopped/exited containers alongside running ones. On CI a stale stopped `gitlab/gitlab-ee` container from a previous failed run would match `FirstOrDefault`, and `RunGitLabRailsRunnerAsync` would then try to exec into it — Docker returns 500 on exec against a stopped container, so all retry attempts would exhaust before throwing.

Drop `All = true` (the default lists only running containers), or add `.Where(c => c.State == "running")` after the call.

Missing trailing comma in inline initializer (project style).

NGitLab.Tests/ProjectsTests.cs
```diff
@@ -355,7 +355,7 @@ public async Task CreateUpdateDelete(bool initiallySetTagsInsteadOfTopics)
         var updatedProject2 = projectClient.Update(createdProject.PathWithNamespace, new ProjectUpdate { Visibility = VisibilityLevel.Internal });
         Assert.That(updatedProject2.VisibilityLevel, Is.EqualTo(VisibilityLevel.Internal));
 
-        projectClient.Delete(createdProject.Id);
+        await projectClient.PermanentlyDeleteAsync(createdProject.Id);
```
msrouchou, line 358: `PermanentlyDeleteAsync` relies on the project being soft-deletable (GitLab ≥ 18.0 behaviour). On a pre-v18 instance `DeleteAsync` destroys the project immediately, so the subsequent `GetAsync` inside the implementation returns 404 and the test fails unconditionally.

Add a version guard:
```csharp
context.IgnoreTestIfGitLabVersionOutOfRange(VersionRange.Parse("[18.0,)"));
```
Same applies to the two calls in `ForkProject` (≈ line 703) and the one in `CreateProjectWithSquashOption` (≈ line 812).

NGitLab.Tests/Docker/GitLabDockerContainer.cs
```diff
@@ -21,26 +21,33 @@ namespace NGitLab.Tests.Docker;
...
     private static readonly SemaphoreSlim s_setupLock = new(initialCount: 1, maxCount: 1);
     private static GitLabDockerContainer s_instance;
 
-    public string Host { get; private set; } = "localhost";
+    /// <summary>
+    /// Set only when the container was spawned locally via Testcontainers.
+    /// On CI, GitLab already runs as a pre-existing service container, so this stays null
+    /// and credential generation falls back to a raw Docker Engine API exec call.
+    /// </summary>
+    private IContainer _localContainer;
```
msrouchou, line 46: `IContainer` is `IAsyncDisposable` but `DisposeAsync` is never called on `_localContainer`. The .NET wrapper leaks its `DockerClient` and HTTP resources each run (`WithReuse` keeps the container process alive, but that's separate). Implement `IAsyncDisposable` on `GitLabDockerContainer` and call `await _localContainer.DisposeAsync()` in the override.

NGitLab.Tests/Docker/GitLabDockerContainer.cs
```diff
@@ -135,151 +140,62 @@ private static async Task ValidateDockerIsEnabled(DockerClient client)
         }
     }
 
-    private async Task SpawnDockerContainerAsync()
+    private async Task SpawnLocalDockerContainerAsync()
```
msrouchou, line 143: The old `SpawnDockerContainerAsync` called `ValidateDockerIsEnabled` first, which gave a clear "Cannot connect to Docker service" message and skipped the test. This method skips that check — a missing Docker daemon now surfaces as a raw Testcontainers exception. Consider calling `ValidateCiDockerIsEnabled` at the top here.

NGitLab.Tests/Docker/GitLabDockerContainer.cs
```diff
@@ -135,151 +140,62 @@ private static async Task ValidateDockerIsEnabled(DockerClient client)
...
 
-            await Task.Delay(5000);
-        }
+        // https://docs.gitlab.com/omnibus/settings/configuration.html
+        // GitLab reports "healthy" long before it's actually ready to serve requests, so we wait
+        // on the HTTP endpoint itself rather than on the container's health status.
+        // WithReuse keeps an existing container running across local test runs (GitLab takes
+        // minutes to boot); bumping LocalGitLabDockerVersion requires removing the old container
+        // manually (`docker rm -f NGitLabClientTests`) since reuse matching is name+config based.
+        _localContainer = new ContainerBuilder(ImageName + ":" + LocalGitLabDockerVersion)
```
msrouchou, line 182: `ContainerBuilder` has no constructor overload taking an image string in Testcontainers 4.x — this won't compile. The fluent API is `new ContainerBuilder().WithImage(ImageName + ":" + LocalGitLabDockerVersion)...`

NGitLab.Tests/Docker/GitLabDockerContainer.cs
```diff
@@ -289,43 +205,45 @@ private async Task GenerateCredentialsAsync()
...
-
             TestContext.Progress.WriteLine("Creating root token via 'gitlab-rails runner'");
 
             // Keep only scopes the running GitLab version supports (an unknown scope makes `create!` raise).
-            const string script = """
+            var script = $"""
                 desired_scopes = %w[api read_user read_api read_repository write_repository sudo admin_mode create_runner manage_runner k8s_proxy]
                 available_scopes = Gitlab::Auth.all_available_scopes.map(&:to_s)
-                token = User.find_by_username!('root').personal_access_tokens.create!(
+                token = User.find_by_username!('{AdminUserName}').personal_access_tokens.create!(
```
msrouchou, line 225: `AdminUserName` is interpolated directly into Ruby source. Safe today since it's hardcoded to `"root"`, but a single quote would break the script. Prefer passing it as an environment variable on the exec call and reading `ENV.fetch('ADMIN_USERNAME')` in Ruby.

NGitLab.Tests/Docker/GitLabDockerContainer.cs
```diff
@@ -361,46 +279,64 @@ void GenerateUserToken()
...
+
+        if (_localContainer is not null)
         {
-            (stdout, stderr) = await stream.ReadOutputToEndAsync(CancellationToken.None).ConfigureAwait(false);
+            var result = await _localContainer.ExecAsync(["gitlab-rails", "runner", script]).ConfigureAwait(false);
+            (stdout, stderr, exitCode) = (result.Stdout, result.Stderr, result.ExitCode);
         }
+        else
+        {
+            var containerId = await ResolveCiGitLabContainerIdAsync(client).ConfigureAwait(false);
```
msrouchou, line 321: `client` is `null` on the local path. The invariant `_localContainer != null ↔ client == null` holds at current call sites, but nothing enforces it — a future caller with both `null` would NRE here. Consider splitting into two private methods (one per path) so each takes only what it needs.

NGitLab.Tests/ProjectsTests.cs
louis-z: I'm not sure it's needed since tests are never run on GitLab < 18.

NGitLab.Tests/Docker/GitLabDockerContainer.cs
louis-z: The version of `Testcontainers` used here is 4.15.0, which does provide a `ContainerBuilder(string image)` ctor. Confirmed by the fact that the project does indeed currently build without errors.

NGitLab.Tests/Docker/GitLabDockerContainer.cs
louis-z: Dropped `All = true`, so only running containers are matched by image name.

NGitLab.Tests/Docker/GitLabDockerContainer.cs
louis-z: `GitLabDockerContainer` now implements `IAsyncDisposable`, disposing `_localContainer` when set. This is wired up via a `[OneTimeTearDown]` in SetUpFixture.cs.

NGitLab.Tests/Docker/GitLabDockerContainer.cs
louis-z: This method now opens a short-lived `DockerClient` and calls `ValidateDockerIsEnabled` _(renamed from `ValidateCiDockerIsEnabled`)_ before building the container, restoring the clear "Cannot connect to Docker service" message.

NGitLab.Tests/Docker/GitLabDockerContainer.cs
louis-z: `AdminUserName` will (most likely) always be hardcoded to "root".

NGitLab.Tests/Docker/GitLabDockerContainer.cs
louis-z: `RunGitLabRailsRunnerAsync(DockerClient, string)` has been split into `RunGitLabRailsRunnerLocallyAsync(string script)` and `RunGitLabRailsRunnerOnCiAsync(DockerClient client, string script)`, sharing exit-code/token parsing via a new `ParseRailsRunnerOutput` helper.

